PR #2, opened 2021-09-30
Setup eslint, prettier, flow, jest

somewhatabstract:
## Summary:
No code yet, but getting code quality tooling in place.

Issue: FEI-4002

## Test plan:
`yarn lint` to see eslint run
`yarn flow` to see flow run
`yarn test` to see jest run

Some will fail because there's nothing for them to check yet.


.eslintrc.js
```diff
@@ -0,0 +1,71 @@
+module.exports = {
+    extends: [
+        "@khanacademy",
+    ],
+    plugins: ["import", "jest", "promise", "monorepo"],
+    settings: {
+        react: {
+            version: "detect",
```
kevinbarabash, line 8: Fancy.  We should start using that in webapp and wonder-blocks as well.

.eslintrc.js
```diff
@@ -0,0 +1,71 @@
...
+        "monorepo/no-relative-import": "error",
+        "import/no-restricted-paths": [
+            "error",
+            {
+                zones: [
+                    {
+                        target: "./packages/(?!.*test.js).*",
+                        from: "utils",
+                    },
+                ],
```
kevinbarabash, line 67: If I'm reading this correctly, we're disallowing importing from the `utils` directory from all packages but excluding test files.  If that's the case, looks good to me.

build-settings/babel.config.js
```diff
@@ -0,0 +1,22 @@
...
+    env: {
+        test: {
+            presets: [
+                "@babel/preset-flow",
+                [
+                    "@babel/preset-env",
+                    {
+                        targets: {
+                            node: true,
+                        },
```
kevinbarabash, line 16: Is there a specific minimum version of node we want to target, e.g. version 12?

build-settings/babel.config.js
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable import/no-commonjs */
+module.exports = {
+    presets: ["@babel/preset-flow", "@babel/preset-env"],
+    plugins: [
+        "@babel/plugin-proposal-class-properties",
+    ],
```
kevinbarabash, line 6: https://babeljs.io/docs/en/babel-plugin-proposal-class-properties says that this plugin is included in @babel/preset-env now so maybe this isn't necessary.

build-settings/package.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "ws-dev-build-settings",
+    "version": "0.0.1",
+    "license": "MIT",
+    "private": true
+}
```
kevinbarabash, line 6: Hopefully this `build-settings` internal package pattern continues to be useful. 🤞

config/jest/test.config.js
```diff
@@ -0,0 +1,29 @@
...
+    moduleNameMapper: {
+        "^@khanacademy/wonder-stuff-(.*)$":
+            "<rootDir>/packages/wonder-stuff-$1/src/index.js",
+    },
+    collectCoverageFrom: [
+        "packages/**/*.js",
+        "!packages/**/*.flowtest.js",
+        "!packages/**/dist/**/*.js",
+        "!<rootDir>/node_modules/",
+        "!packages/**/node_modules/",
```
kevinbarabash, line 24: Instead of collecting coverage from `packages/**/*.js` and then having to exclude node_modules, we should probably collect coverage on `packages/*/src/**/*.js`.  Having all of the source files live in `packages/*/src` will also make migrating to TypeScript easier because it makes using project reference and incremental compilation easier because `out/` can be a sibling of `src/`.  In this situation `out/` is separate from `dist/` and acts more like the babel cache.

config/jest/test.config.js
```diff
@@ -0,0 +1,29 @@
...
+ * test environment: jest-environment-node.
+ */
+const path = require("path");
+
+module.exports = {
+    rootDir: path.join(__dirname, "../../"),
+    transform: {
+        "^.+\\.jsx?$": "<rootDir>/config/jest/test.transform.js",
+    },
+    testEnvironment: "jest-environment-node",
```
kevinbarabash, line 12: Good defaults are good.

package.json
```diff
@@ -2,9 +2,42 @@
...
+    "flow-bin": "^0.161.0",
+    "jest": "^27.2.4",
+    "jest-extended": "^0.11.5",
+    "prettier": "^2.4.1"
+  },
+  "scripts": {
+    "clean": "rm -rf packages/wonder-stuff-*/dist && rm -rf packages/wonder-stuff-*/node_modules",
+    "coverage": "yarn run jest --coverage",
+    "flow-ci": "flow check --max-workers 0",
+    "jest": "jest --config config/jest/test.config.js",
```
kevinbarabash, line 39: Not that it makes a difference, but the `--config` isn't necessary since there's a jest.config.js in the project root.

package.json
```diff
@@ -2,9 +2,42 @@
...
+    "lint": "eslint --config .eslintrc.js packages",
```
kevinbarabash, line 40: Similar, eslint should recognize the .eslintrc.js without the `--config` option.

kevinbarabash (changes requested):
Thanks for setting all of these tools up.  Requesting changes to `collectCoverageFrom` in the jest config.  See inline comments for details.

jeresig (approved):
Looks ok to me!

config/jest/test.config.js
kevinbarabash: I see we're putting source files in `packages/*/src` in #4 so this is no longer blocking but more of possible tweak.

kevinbarabash (approved):
Changes in #4 reduced the severity of the issue I raised in the PR.  :shipit:

PR #4, opened 2021-09-30
Add first package, rollup build, and fix jest config

somewhatabstract:
## Summary:
Adds `wonder-stuff-core` package and adds rollup build for esm.

Also, it turns out jest and babel-jest changed a few things since the version Wonder Blocks is using, so this fixes that.

Issue: FEI-4002

## Test plan:
`yarn test`
`yarn flow`
`yarn lint`

Also tried wallaby, which works OTB without a wallaby.js config.

We should see the Github Actions passing now.


jeresig (approved):

packages/wonder-stuff-core/src/index.js
```diff
@@ -0,0 +1,3 @@
+// @flow
+// eslint-disable-next-line no-console
+console.log("Hello World!");
```
kevinbarabash, line 3: Thank you for having source live inside `src`. ❤️

build-settings/rollup.config.js
```diff
@@ -0,0 +1,29 @@
...
+        file: `packages/${pkgName}/dist/index.js`,
+        format: "esm",
+    },
+    input: `packages/${pkgName}/src/index.js`,
+    plugins: [
+        babel({
+            babelHelpers: "bundled",
+            presets,
+            plugins,
+            exclude: "node_modules/**",
```
kevinbarabash, line 20: I assume this also works for imports of other `ws-*` modules from one another since we aren't doing any module mapping shenanigans here.

.vscode/settings.json
```diff
@@ -6,13 +6,9 @@
     "typescript.format.enable": false,
     "typescript.validate.enable": false,
     "javascript.validate.enable": false,
-    "coverage-gutters.lcovname": "./coverage/lcov.info",
-    "jest.enableSnapshotUpdateMessages": true,
-    "jest.pathToConfig": "./jest.config.js",
```
kevinbarabash, line 11: Is the plan to replace this with a wallaby config eventually?

.eslintrc.js
```diff
@@ -2,10 +2,13 @@ module.exports = {
...
-    plugins: ["import", "jest", "promise", "monorepo"],
+    plugins: ["import", "jest", "promise", "monorepo", "disable"],
     settings: {
-        react: {
-            version: "detect",
+        "eslint-plugin-disable": {
+            paths: {
+                react: ["./*.js", "src/*.js"],
+                "jsx-a11y": ["./*.js", "src/*.js"],
+            },
```
kevinbarabash, line 11: What's adding rules for `react` and `jsx-a11y` that we have to disable them here?

kevinbarabash (approved):
LGTM.  A few questions, but nothing blocking.  Thanks for setting up the build tooling.

build-settings/rollup.config.js
somewhatabstract: I think so. This is based on WB config

PR #11, opened 2021-10-08
Add support for browser and node outputs from a package

somewhatabstract:
## Summary:
Also:
- Add support for process.env.NODE_ENV and package watching
- Update prepublish checks to cover browser setting

Overall, this gets us setup to publish packages that support ESM and CJS
in browser and node. We can probably drop the browser CJS file if we
really wanted, but this way, it's safe - no one can accidentally include
the wrong thing.

Also includes support for differing behavior based on the value of
`process.env.NODE_ENV` during the build. Downside here of course is that
if code wanted to differentiate based on the value of `process.env.NODE_ENV`
at runtime, they will have to use a different syntax to get the value so
that it isn't replaced at build time. This is entirely possible though,
so I think that's just fine.

I believe this concludes our initial setup of the repo. Time to add code.

Issue: FEI-4002

## Test plan:
`yarn flow`
`yarn test`
`yarn build`
`yarn build --configPackages=core`
`yarn build --configPackages=poop` fails as it should
`yarn watch`
`yarn watch --configPackages=core`
`yarn watch --configPackages=poop` fails as it should
`node ./utils/pre-publish-check.js` against various edits of the core package.json to verify the checks are working.

### DIFF
--- a/build-settings/rollup.config.js
+++ b/build-settings/rollup.config.js
@@ -1,33 +1,47 @@
 /* eslint-disable import/no-commonjs */
 import fs from "fs";
+import path from "path";
 import autoExternal from "rollup-plugin-auto-external";
 import {babel} from "@rollup/plugin-babel";
 import {terser} from "rollup-plugin-terser";
 import copy from "rollup-plugin-copy";
+import resolve from "@rollup/plugin-node-resolve";
 
-const {presets, plugins} = require("./babel.config.js");
+const createBabelPresets = require("./create-babel-presets.js");
 
 /**
- * Creates the rollup config shared by all our output formats.
+ * Make path to a package relative path.
  */
-const createSharedConfig = (pkgName) => ({
-    input: `packages/${pkgName}/src/index.js`,
+const makePackageBasedPath = (pkgName, pkgRelPath) =>
+    path.normalize(path.join("packages", pkgName, pkgRelPath));
+
+/**
+ * Generate the rollup output configuration for a given
+ */
+const createOutputConfig = (pkgName, format, targetFile) => ({
+    file: makePackageBasedPath(pkgName, targetFile),
+    sourcemap: true,
+    format,
+});
+
+const createConfig = ({name, format, platform, file}) => ({
+    output: createOutputConfig(name, format, file),
+    input: makePackageBasedPath(name, "./src/index.js"),
     plugins: [
         babel({
             babelHelpers: "bundled",
-            presets,
-            plugins,
+            presets: createBabelPresets({platform, format}),
             exclude: "node_modules/**",
         }),
+        resolve({
+            browser: platform === "browser",
+        }),
         autoExternal({
-            packagePath: `packages/${pkgName}/package.json`,
+            packagePath: makePackageBasedPath(name, "./package.json"),
         }),
         terser(),
 
         // This generates the flow import file.
-        // The copyOnce ensures that we copy just once for a specific target
-        // file, even if we have multiple copy plugin configs for that file,
-        // so it's ok that we duplicate this step in our two outputs.
         copy({
             copyOnce: true,
             verbose: true,
@@ -37,7 +51,7 @@ const createSharedConfig = (pkgName) => ({
                     // with ./
                     src: "build-settings/index.flow.js.template",
                     // dest path is relative to src path.
-                    dest: `packages/${pkgName}/dist`,
+                    dest: makePackageBasedPath(name, "./dist"),
                     rename: "index.flow.js",
                 },
             ],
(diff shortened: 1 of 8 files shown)


build-settings/rollup.config.js
```diff
@@ -1,33 +1,56 @@
...
+    input: makePackageBasedPath(name, "./src/index.js"),
     plugins: [
+        replace({
+            preventAssignment: true,
+            values: {
+                "process.env.NODE_ENV": JSON.stringify(
+                    process.env.NODE_ENV || "production",
+                ),
+            },
+        }),
```
somewhatabstract, line 39: This enables us to do `process.env.NODE_ENV` checks during build; and rollup will omit code accordingly.

build-settings/rollup.config.js
```diff
@@ -1,33 +1,56 @@
...
         babel({
             babelHelpers: "bundled",
-            presets,
-            plugins,
+            presets: createBabelPresets({platform, format}),
             exclude: "node_modules/**",
         }),
+        resolve({
+            browser: platform === "browser",
+        }),
```
somewhatabstract, line 47: This tells the module resolution to use the `browser` field of package.json when it exists, but only for browser builds.

build-settings/rollup.config.js
```diff
@@ -37,49 +60,76 @@ const createSharedConfig = (pkgName) => ({
...
+ * packages.
+ */
+const createRollupConfig = (commandLineArgs) => {
+    const {configPackages} = commandLineArgs;
+    const pkgNames =
+        configPackages && configPackages.length
+            ? configPackages
+                  .split(",")
+                  .map((s) =>
+                      s.startsWith("wonder-stuff") ? s : `wonder-stuff-${s}`,
```
somewhatabstract, line 128: This bit means we can target rollup at specific packages if we want.

package.json
```diff
@@ -43,7 +45,8 @@
     "rollup-plugin-terser": "^7.0.2"
   },
   "scripts": {
-    "build": "rollup -c build-settings/rollup.config.js",
+    "build": "NODE_ENV=production rollup -c build-settings/rollup.config.js",
+    "watch": "NODE_ENV=development rollup -c build-settings/rollup.config.js --watch",
```
somewhatabstract, line 49: Added a `watch` option as I think it will be useful - such as if I'm using Quokka to test the actual package output and want to tweak things as I go.

package.json
```diff
@@ -43,7 +45,8 @@
     "rollup-plugin-terser": "^7.0.2"
   },
   "scripts": {
-    "build": "rollup -c build-settings/rollup.config.js",
+    "build": "NODE_ENV=production rollup -c build-settings/rollup.config.js",
```
somewhatabstract, line 48: Explicitly set things to production here just to be explicit.

packages/wonder-stuff-core/package.json
```diff
@@ -4,7 +4,7 @@
     },
     "private": true,
     "name": "@khanacademy/wonder-stuff-core",
-    "version": "1.0.0",
+    "version": "0.0.1",
```
somewhatabstract, line 7: Set to 0.0.1 in response to @kevinbarabash's comment on an earlier PR. :)

somewhatabstract: Maybe it should be 0.0.0 so lerna will make it 0.0.1 on first release? 🤷🏻

packages/wonder-stuff-core/package.json
```diff
@@ -15,6 +15,10 @@
     "devDependencies": {
         "ws-dev-build-settings": "^0.0.2"
     },
+    "browser": {
+        "dist/es/index.js": "dist/es/index.browser.js",
+        "dist/index.js": "dist/index.browser.js"
+    },
```
somewhatabstract, line 21: Allows me to test at the new build stuff.

packages/wonder-stuff-core/src/index.js
```diff
@@ -1,3 +1,16 @@
...
+        break;
+
+    case "development":
+        console.log("Hello development World!");
+        break;
+
+    case "test":
+        console.log("Hello test World!");
+        break;
+}
```
somewhatabstract, line 16: Allows me to test out the `process.env.NODE_ENV` support and see how it impacts the output.

build-settings/rollup.config.js
jeresig: Huh - rollup doesn't support this natively? That's surprising!

build-settings/rollup.config.js
jeresig: Is there any way to make this generic? (e.g. not reference the string `wonder-stuff` directly?) If so we could re-use this rollup config in a bunch of other places, including Wonder Blocks and maybe Perseus!

jeresig (approved):
This looks great! Is there any way for us to make all this monorepo/build config stuff generic so that we can share it between wonder blocks, wonder stuff, and perseus?

build-settings/rollup.config.js
jeresig: Ideally I'd like to avoid copy-and-pasting configs as it will mean that things will get quickly out-of-sync, so anything we can do to make it generic and easily shareable would be ideal!

build-settings/rollup.config.js
somewhatabstract: Yeah. I based this off React's use of rollup. I was a little surprised too.

PR #16, opened 2021-10-12
Implement KindError

somewhatabstract:
## Summary:
This implements `KindError`, an error type to act as the basis for our various `KAError` versions.
This includes code to combine messages and stack traces in a meaningful way between a source error and the error being constructed around that.

Also:

 - Enable no-default-exports lint rule
   - Named exports provide an explicit API that is easy to search, rename, etc.
   - Required updating to @babel/eslint-parser
 - Excludes index.js and types.js from coverage data
 - Adds jest-when

This doesn't add any of the useful helper methods from mobile yet or anything sentry specific.
The sentry stuff will be in a different package than core and the helpers will be added in an upcoming PR (either to KindError or a more specific derivation).
Some of the helpers should be unnecessary now as the constructor uses an `options` arg to simplify the provision of various options.

Issue: FEI-3997

## Test plan:
`yarn test`
`yarn flow`
`yarn lint`
`yarn build`

I also want to create some sort of E2E/integration test (see FEI-4058 for details on that)


packages/wonder-stuff-core/src/kind-error.js
```diff
@@ -0,0 +1,125 @@
...
+            metadata,
+            stripStackFrames,
+            minimumFrameCount,
+        }: Options = {
+            cause: undefined,
+            metadata: undefined,
+            prefix: "",
+            stripStackFrames: 0,
+            minimumFrameCount: 1,
+        },
```
somewhatabstract, line 62: By using `Options`, it should make this much easier to build without needing helpers like `withSourceError` that the current `KAError` in mobile has added (which are great, yet feel like a workaround for the constructor lacking options like this).

packages/wonder-stuff-core/src/kind-error.js
```diff
@@ -0,0 +1,125 @@
...
+             * stack, but I don't want this crashing because we try to
+             * assign to a readonly thing, so being mildly cautious.
+             */
+            delete this.stack;
+            this.stack = combined.toString();
+
+            // We update our message to a normalized one from the combined
+            // stacks, as this gives us a nice causal relationship.
+            this.message = combined.message;
+        }
```
somewhatabstract, line 123: This is new art based on the old ways.
1. webapp would just swap the KAError stack trace for the source error stack trace
2. mobile would concat the stacks together, with the source error first and the KAError stack second
3. Now, in this version, we do some work to combine the stacks in a chronological manner that hopefully retains value but reduces repetition. See `ErrorInfo.combine` for details.

packages/wonder-stuff-core/src/kind-error.js
lillialexis: Also, you don't need the builder methods. You can always still use the constructor, but they're nice-to-haves, because I found that using the constructor, and thinking through the constructor args, was just taking too much effort, even though that extra effort was really tiny. Tiny was still too much, lol.

somewhatabstract: That's why I used an options type rather than individual args. Provides the same level of discovery. There's nothing stopping derivations adding more things like helpers.

packages/wonder-stuff-core/src/kind-error.js
```diff
@@ -0,0 +1,125 @@
...
+    +metadata: ?$ReadOnly<Metadata>;
+
+    /**
+     * Creates an instance of `KindError`.
+     *
+     * @param {string} message The error message.
+     * @param {TKinds} [kind] The kind of error. This will be combined with
+     * `prefix` to form the name of the error, i.e. PrefixKindError.
+     * Defaults to `Errors.Unknown`.
+     * @param {Error} [cause] The error that caused this error.
```
lillialexis, line 36: Is the docstring syntax right here? Since you have an object, `Options`, as an arg, not the individual properties of the `Options` as args. I think for correct docstrings, you'd want to define the Options type, and doc it's props where you define the type.

packages/wonder-stuff-core/src/kind-error.js
lillialexis: I like the `Options` because then you don't have to worry about getting the order of the constructor arguments correct, and it supports multiple args being `null` and some not being `null`, without default values or other weirdnesses coming in to play. In that sense, it's totally solved half the problem!

It doesn't solve the problem of the developer having to discover the shape of what could be a very large object. The experience in my IDE might not be a nice, as the IDE attempts to suggest to me the arguments in-line as I type, because now they're down one level.

I'm not saying that we change _this_, just that we also still want builder methods!

packages/wonder-stuff-core/src/kind-error.js
lillialexis: It might be fine... I'm not sure any more what kinds of fancy stuff is expected by documentation generators and IDEs.

somewhatabstract: I haven't updated the docstring yet since I did the `Options` stuff.

packages/wonder-stuff-core/src/error-info.js
```diff
@@ -0,0 +1,200 @@
...
+     *        caused by
+     *            Error: A
+     *      stackA2
+     *      stackA1
+     *      stackB2
+     *      stackB1
+     *      stackShared2
+     *      stackShared1
+     */
+    static combine(consequence: ErrorInfo, cause: ErrorInfo): ErrorInfo {
```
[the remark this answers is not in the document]
somewhatabstract: Naming things is hard. This has changed names a lot. Perhaps `include`?

packages/wonder-stuff-core/src/kind-error.js
lillialexis: Oh wait, maybe context is too undefined... wonder what `Tags` looks like...

somewhatabstract: Even if you type something like...

```
{
    "
}
```

Shame.

lillialexis: But because `SentryTags` object is a defined flow-type, while in the auto-complete for `withTags`, I can click through to the definition/docs of `SentryTags`! I don't think it works on the docstrings for the individual properties though.
Screen Shot 2021-10-14 at 2 44 44 PM

packages/wonder-stuff-core/src/clone-metadata.js
```diff
@@ -0,0 +1,39 @@
...
+        return metadata;
+    }
+
+    // Clone the data.
+    const clone = cloneValue(metadata);
+
+    // We know that we have cloned the incoming data and so our clone is of
+    // the correct type.
+    // $FlowIgnore[incompatible-return]
+    return Object.freeze(clone);
```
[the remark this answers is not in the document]
somewhatabstract: Read-only in and read-only out. Unless there's a reason not to, I want things to remain immutable.

packages/wonder-stuff-core/src/error-info.js
```diff
@@ -0,0 +1,200 @@
...
+        // mutate outside our control.
+        return [...this._stackFrames];
+    }
+
+    /**
+     * Get the stack trace as a string.
+     */
+    toString(): string {
+        // Now construct the string value.
+        return `${this._message}\n${this._stackFrames.join("\n")}`;
```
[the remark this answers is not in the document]
somewhatabstract: This is doing the same thing as what currently happens.

packages/wonder-stuff-core/src/clone-metadata.js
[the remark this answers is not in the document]
lillialexis: Right on, fair enough. (Though I can think of a few use cases where maybe we'd want to update this? Mostly just in working around the already-existing mobile code... sometimes errors get passed around and updated and _then_ logged.)

packages/wonder-stuff-core/src/error-info.js
```diff
@@ -0,0 +1,200 @@
...
+     * This method would produce the equivalent of:
+     *    Error: B
+     *        caused by
+     *            Error: A
+     *      stackA2
+     *      stackA1
+     *      stackB2
+     *      stackB1
+     *      stackShared2
+     *      stackShared1
```
lillialexis, line 86: oooooh, I understand now why this method is also doing the message-combining too! It's because the calls stacks of errors start with the error message on the first frame!

Hmmm... I think they way Java does it is more like
```
Error: B 
   frame
   frame
caused by
Error A
    frame
    frame
```

I don't really know if one way is better than the other or easier than the other... especially if we go n-deep on error nesting...

But this and the fact that `toString()` is the error message only, on mobile, I'll think more on this.

Anyway, this is only the first diff! Feel free to land and increment changes! :-)

packages/wonder-stuff-core/src/error-info.js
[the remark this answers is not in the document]
somewhatabstract: I renamed it to `fromConsequenceAndCause` which sounds like an homage to Jane Austen. 😂

packages/wonder-stuff-core/src/kind-error.js
somewhatabstract: Updated the docstring now.

packages/wonder-stuff-core/src/error-info.js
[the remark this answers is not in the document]
somewhatabstract: This is in the `ErrorInfo` class, not in the `KindError`. I don't think this will cause any issues.

packages/wonder-stuff-core/src/error-info.js
somewhatabstract: Yeah, I'm looking to construct a stack that matches JavaScript platforms, which tend to expect a message on the first line, but then stack frames for the rest. And this is only used in the case of adding a cause to an error, so it doesn't affect the other possible things we will be building messages for.

somewhatabstract: Even with nesting, this will all work as the same rules will apply and, just as with the mobile code now, the combining is done during error construction and the same rules still apply.

packages/wonder-stuff-core/src/error-info.js
```diff
@@ -0,0 +1,203 @@
...
+     */
+    static fromConsequenceAndCause(
```
lillialexis, line 88: **_LOVE!!_** 📖 😻

somewhatabstract: I guess the Jane Austen version would be "Cause and Consequence" 🤣

PR #20, opened 2021-10-15
Implement KindSentryError basics

somewhatabstract:
## Summary:
This adds the `wonder-stuff-sentry` package (I expect more sentry specific stuff to come and it really isn't needed by things that might include core).
This includes the `KindSentryError` which adds some more of the things we need for sentry; specifically, `SentryData` and some helpers to merge that data from causal errors.

This doesn't yet include the other things for sentry like `groupByMessage` or the helpers that turn this into actual sentry API calls.
This also doesn't include any of the nice builder methods yet. The intent is to start adding this stuff in the next PR - however, since different platforms will call the Sentry API from different package imports, I will likely use a multiple package approach - one for node and one for browser, so that only the code needed is included. Something like `wonder-stuff-sentry-node` and `wonder-stuff-sentry-browser`. The intent will be for `wonder-stuff-sentry` to detect which one is in use, and use it accordingly so that consuming tools just need to install the appropriate package and then call things. This will allow those other packages to then have the relevent Sentry piece as a peer dependency of the correct packages.

Issue: FEI-3997

## Test plan:
`yarn test`
`yarn build`
`yarn flow`


build-settings/rollup.config.js
```diff
@@ -58,10 +58,10 @@ const createConfig = ({name, format, platform, file}) => ({
                 {
                     // src path is relative to the package root unless started
                     // with ./
-                    src: "build-settings/index.flow.js.template",
+                    src: "build-settings/index.js.flow.template",
                     // dest path is relative to src path.
                     dest: makePackageBasedPath(name, "./dist"),
-                    rename: "index.flow.js",
+                    rename: "index.js.flow",
```
somewhatabstract, line 64: This was a typo. As soon as I started trying to get flow working across packages, I realised my error.

flow-typed/npm/jest-when.js
```diff
@@ -0,0 +1,28 @@
+// NOTE(@somewhatabstract): Crafted from looking at the TypeScript types and
+// editing to match the Jest flow types for maintenance clarity.
```
somewhatabstract, line 2: Created these types because I think they're useful and otherwise, we'd have to import `jest-when` as `any`-typed, which I want to clamp down on in this new repo.

packages/wonder-stuff-core/src/__tests__/error-info.test.js
```diff
@@ -58,6 +58,7 @@ describe("ErrorInfo", () => {
 
             // Act
             const act = () =>
+                // $FlowIgnore[unclear-type]
```
somewhatabstract, line 61: Turned on linting against `Object`, `any`, and `Function`. `any` is the only one where we sometimes need it, as far as I can tell - and usually just for casting things in tests or for initializing empty objects.

I think that here in the test, we could just suppress the error and not cast at all, so I'll probably change this and related locations.

packages/wonder-stuff-core/src/kind-error.js
```diff
@@ -104,13 +109,13 @@ export class KindError extends Error {
         if (kind && /\s/g.test(kind)) {
             throw new Error("kind must not contain whitespace");
         }
-        if (prefix && /\s/g.test(prefix)) {
+        if (prefix != null && /\s/g.test(prefix)) {
```
somewhatabstract, line 112: Linting against sketchy null and sketchy number caught a few things that we weren't handling right. So this fixes them up.

packages/wonder-stuff-core/src/types.js
```diff
@@ -5,9 +5,10 @@ type MetadataPrimitive = string | number | boolean;
  * A collection of data.
  */
 export type Metadata = {
-    +[name: string]: ?(
+    [name: string]: ?(
```
somewhatabstract, line 8: No need for `Metadata` to be directly read-only up front.

packages/wonder-stuff-sentry/src/__tests__/kind-sentry-error.test.js
```diff
@@ -0,0 +1,417 @@
+// @flow
+import {KindError, Errors} from "@khanacademy/wonder-stuff-core";
```
somewhatabstract, line 2: Will need the github action to build locally, I think (and maybe call `lerna bootstrap` if we want this to pass before any package is published.

packages/wonder-stuff-sentry/src/kind-sentry-error.js
```diff
@@ -0,0 +1,170 @@
...
+     * @type {?Error}
+     */
+    cause?: ?Error,
+
+    /**
+     * Data to be attached to the error that would be reported to Sentry.
+     *
+     * @type {?$ReadOnly<$Partial<SentryData>>}
+     */
+    sentryData?: ?$ReadOnly<$Partial<SentryData>>,
```
somewhatabstract, line 29: Using the new Flow `$Partial` type :smile:

somewhatabstract: This allows callers to only specify a portion of the expected sentry info, even though we always assert a complete set (even if they are empty objects) elsewhere. I did this because I feel that it makes code easier to work with if we don't need null checks first.

packages/wonder-stuff-sentry/src/kind-sentry-error.js
```diff
@@ -0,0 +1,170 @@
...
+     * all the SentryData from this and causal errors it contains.
+     */
+    getMergedSentryData(): $ReadOnly<SentryData> {
+        // First, get the stack of data from all causal errors and this one.
+        const sentryDataArray = this.getAllSentryData();
+
+        // Then, reduce the stack of data into a single object.
+        // Note that we use `reduceRight` so that the most recent data is at the
+        // top of the stack and takes precedence over the older data.
+        return sentryDataArray.reduceRight(sentryDataReducer, EmptySentryData);
```
somewhatabstract, line 168: @lilli this plus the reducer does the same as the equivalent code and util methods in the mobile repo.

`reduceRight` reduces from the end of the array to the start, as we want.

packages/wonder-stuff-core/src/kind-error.js
```diff
@@ -124,19 +129,21 @@ export class KindError extends Error {
 
         // Set the name so we get a nice error output, like
         // KAInternalError
-        this.name = `${prefix ?? ""}${kind}Error`;
+        this.name = `${prefix ?? ""}${kind}${name ?? "Error"}`;
```
somewhatabstract, line 132: Added the ability to specify the name of the error, so that we can have `SentryError` at the end for Sentry error stuff.

So I'm going to add this to the `SentryError` too, so that it can be overridden.
This will change what gets recorded in Sentry, so I am pondering on whete

packages/wonder-stuff-sentry/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "dependencies": {
+        "@khanacademy/wonder-stuff-core": "^0.0.1"
+    },
+    "devDependencies": {
+        "ws-dev-build-settings": "^0.0.2"
+    },
+    "browser": {
+        "dist/es/index.js": "dist/es/index.browser.js",
+        "dist/index.js": "dist/index.browser.js"
+    },
```
kevinbarabash, line 23: TIL, neat 🙂

packages/wonder-stuff-sentry/src/__tests__/kind-sentry-error.test.js
kevinbarabash: wonder-blocks uses `import/no-unresolved` as well.  `yarn test` is set up to build things first before it runs lint.  You may also have to do `lerna bootstrap` as well as you mentioned.

packages/wonder-stuff-core/src/__tests__/kind-error.test.js
```diff
@@ -52,6 +52,42 @@ describe("KindError", () => {
...
+                const act = () =>
+                    new KindError("MESSAGE", Errors.Unknown, {name});
+
+                // Assert
+                expect(act).toThrowErrorMatchingSnapshot();
+            },
+        );
+
         it.each(["K I N D", "KI\nND"])(
             "should throw if the kind has whitespace like %s",
```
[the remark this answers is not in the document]
somewhatabstract: Duplicate of which one?

packages/wonder-stuff-core/src/__tests__/error-info.test.js
kevinbarabash: My bad.  I was looking at the line below.  
```
ErrorInfo.fromConsequenceAndCause((null: any), cause);
```
Why doesn't the arg type for `ErrorInfo.fromConsequenceAndCause` include `null`?

somewhatabstract: Because we don't want folks to call it this way, but we have no control over how it gets called. Just because flow asserts things doesn't mean our consumer will be using flow or not have a bug that passes a bad value.

packages/wonder-stuff-core/src/__tests__/kind-error.test.js
```diff
@@ -240,6 +273,30 @@ describe("KindError", () => {
...
+                });
+
+                // Assert
+                expect(result.message).toMatchInlineSnapshot(`
+                    "UnknownError: ROOT_MESSAGE
+                    	caused by
+                    		UnknownError: UnknownError: CAUSE_2_MESSAGE
+                    	caused by
+                    		Error: CAUSE_MESSAGE"
+                `);
```
somewhatabstract, line 298: @lillialexis here's the test I was talking about :)

packages/wonder-stuff-sentry/src/__tests__/kind-sentry-error.test.js
```diff
@@ -0,0 +1,429 @@
...
+            expect(result).toIncludeAllMembers([
+                {
+                    tags: {},
+                    contexts: {},
+                    fingerprint: ["a", "b", "c"],
+                },
+            ]);
+        });
+
+        it("should return a stack with SentryData from the error and all nested KindSentryErrors, skipping KindErrors", () => {
```
kevinbarabash, line 252: Cool!

packages/wonder-stuff-sentry/src/__tests__/kind-sentry-error.test.js
```diff
@@ -0,0 +1,429 @@
...
+                    },
+                    contexts: {},
+                    fingerprint: [],
+                },
+            ]);
+        });
+    });
+
+    describe("#getMergedSentryData", () => {
+        it("should reduce sentry data stack from bottom to top", () => {
```
kevinbarabash, line 315: I'm not sure how important it is to test this implementation detail as long as the output is correct.  No action req'd.

packages/wonder-stuff-sentry/src/__tests__/sentry-data-reducer.test.js
```diff
@@ -0,0 +1,186 @@
...
+        // Assert
+        expect(result).toEqual({
+            tags: {
+                tag1: "value1Current",
+                tag2: "value2Accumulator",
+                tag3: "value3Current",
+            },
+            contexts: {
+                context1: {
+                    value1: "value1Current",
```
kevinbarabash, line 142: The values you picked make it really easy to see what's going on in these tests.

packages/wonder-stuff-sentry/src/sentry-data-reducer.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+export const sentryDataReducer = (
+    accumulator: SentryData,
+    current: ?$ReadOnly<SentryData>,
+    index: number,
+    array: $ReadOnlyArray<?SentryData>,
+): SentryData => {
+    if (current == null) {
+        return accumulator;
+    }
```
kevinbarabash, line 27: You're programming style is very defensive.  I could definitely stand to adopt some of these practices. 🙂

packages/wonder-stuff-sentry/src/types.js
```diff
@@ -0,0 +1,66 @@
...
+ * @property {?$ReadOnly<SentryContexts>} contexts The contexts to add for the associated
+ * sentry event.
+ * This is equivalent to using the `setContext` API.
+ * NOTE: `setExtras` is deprecated. For similar behavior, use a context keyed as
+ * `"Additional Data"`.
+ * @property {?$ReadOnlyArray<string>} fingerprint The fingerprint of the
+ * associated sentry event.
+ * This is equivalent to using the `setFingerprint` API.
+ */
+export type SentryData = {|
```
kevinbarabash, line 46: This is the only type that's used outside of the file.  Exporting the other ones doesn't hurt and could come in handing the future.

kevinbarabash (approved):
LGTM.  You may want someone with more sentry knowledge to review the sentry specific details.  The organization of things into KindError and KindSentryError makes sense to me.

I'm curious about the use of the word `Kind` in this context.  Is it "kind" in the sense of nice/forgiving/etc. or does is it a stand in for type/category/etc.?

packages/wonder-stuff-core/src/__tests__/kind-error.test.js
```diff
@@ -240,6 +273,30 @@ describe("KindError", () => {
...
+                // Arrange
+                const cause1 = new Error("CAUSE_MESSAGE");
+                const cause2 = new KindError(
+                    "CAUSE_2_MESSAGE",
+                    Errors.Unknown,
+                    {cause: cause1},
+                );
+
+                // Act
+                const result = new KindError("ROOT_MESSAGE", Errors.Unknown, {
```
lillialexis, line 287: I feel like `ROOT` implies this is at the bottom of the nesting, not the top.

packages/wonder-stuff-core/src/__tests__/kind-error.test.js
lillialexis: yay!

packages/wonder-stuff-core/src/__tests__/kind-error.test.js
```diff
@@ -240,6 +273,30 @@ describe("KindError", () => {
...
+                    cause: cause2,
+                });
+
+                // Assert
+                expect(result.message).toMatchInlineSnapshot(`
+                    "UnknownError: ROOT_MESSAGE
```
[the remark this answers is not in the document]
lillialexis: Though, they are in different font-sizes/boldnesses, so it's not _quite_ so awkward...

packages/wonder-stuff-sentry/src/kind-sentry-error.js
```diff
@@ -0,0 +1,177 @@
...
+const EmptySentryData: $ReadOnly<SentryData> = Object.freeze({
+    tags: {},
+    contexts: {},
+    fingerprint: [],
+});
+
+/**
+ * Options for constructing a `KindError`.
+ */
+type Options = {|
```
lillialexis, line 16: Do we need to duplicate this object here? Or can we just spread `kind-errors`'s into this one and add the Sentry data?

packages/wonder-stuff-sentry/src/kind-sentry-error.js
```diff
@@ -0,0 +1,177 @@
...
+    /**
+     * An error responsible for the error being created.
+     *
+     * @type {?Error}
+     */
+    cause?: ?Error,
```
lillialexis, line 22: Tiniest nit: I don't know why, but I'm feeling split on calling this `cause` versus `sourceError`. I can't explain why I like one better than the other, it's just a feeling thing. But for whatever reason, the word `cause` just makes my brain have to do, like, an extra two seconds of processing. I mean, it makes sense to use "cause" because of the language "caused by", but maybe it's just an iota too vague and overloaded, while "sourceError" is just fully unambiguous.

packages/wonder-stuff-sentry/src/kind-sentry-error.js
```diff
@@ -0,0 +1,177 @@
...
+        // Option 2 may make more readable data when stringified, but needs
+        // a policy on what to do about a key collision.
+        // For simplicity of implementation and ease of API use, we choose
+        // option 1.
+        super(message, kind, {
+            ...restOptions,
+            name: name ?? "Sentry",
+            metadata: {
+                sentry: {
+                    // We set the defaults here so that we know these will be
```
lillialexis, line 103: ~SentryData is _next to_ Metadata in the Options arg, but then in here it's added to the `metadata` object. That feels weird to me? Not saying that it needs to change, but just that it feels weird... maybe I just don't understand the code well enough yet...~

OMG OMG OMG, totally missed the above comment 🤦‍♀️

lillialexis: I mean, it does still feel weird though... maybe we don't need to add SentryData to metadata? Like, once in-use, if it's a SentryError, it's a SentryError? For typing and stuff too... I dunno, can't explain, but feels like an anti-pattern.

lillialexis: Another option is to give the super-class, kind-error a "void-star" like pointer to something arbitrary that all sub-classes can use. It's kind of like exactly what you're doing with `Metadata`, but instead of having `SentryData` as a child of `Metadata`, it just _is_ `Metadata`.

packages/wonder-stuff-sentry/src/kind-sentry-error.js
lillialexis: sweet 🍬

packages/wonder-stuff-sentry/src/__tests__/sentry-data-reducer.test.js
lillialexis: Agreed!!

Maybe worth checking all the unit tests in the mobile code base to make sure lots of combos are covered?

packages/wonder-stuff-sentry/src/kind-sentry-error.js
```diff
@@ -0,0 +1,177 @@
...
+ * @class KindSentryError
+ * @extends {KindError}
+ */
+export class KindSentryError extends KindError {
+    constructor(
+        message: string,
+        kind: string = Errors.Unknown,
+        // It's perfectly valid for options to be an empty object.
+        // $FlowIgnore[unclear-type]
+        options: Options = ({}: any),
```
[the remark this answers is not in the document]
somewhatabstract: Hmm, possibly. I'll try it.

packages/wonder-stuff-core/src/__tests__/kind-error.test.js
somewhatabstract: Hmm, I was thinking of it like with directory structures. 🤷🏻

packages/wonder-stuff-sentry/src/types.js
```diff
@@ -0,0 +1,66 @@
+// @flow
+import type {Metadata} from "@khanacademy/wonder-stuff-core";
+
+/**
+ * Tags for a Sentry event.
+ *
+ * The key of each item is the name of the tag.
+ */
+export type SentryTags = {
```
lillialexis, line 9: Even though the Sentry API allows for arbitrary tags, I locked this down to a set list of known-tags for a few reasons:
- Adding tags means they're always there in our Sentry project, and I wanted to avoid "Tag Proliferation"
- Avoids misspellings (which leads to "Tag Proliferation" and confusion on which is the right tag to look at when in the Sentry dashboard trying to find issues and stuff)
- Makes the types for each Tag clear to the developer
- Allows you to search Tags _across_ issues

packages/wonder-stuff-core/src/kind-error.js
```diff
@@ -124,21 +132,25 @@ export class KindError extends Error {
...
             const normalizedError = ErrorInfo.normalize(
                 this,
-                stripStackFrames,
-                minimumFrameCount,
+                stripStackFrames ?? 0,
+                minimumFrameCount ?? 1,
             );
-            const normalizedCause = ErrorInfo.normalize(cause);
+            // We don't normalize the cause as this should have happened already
+            // when it was, itself, created with a cause.
```
lillialexis, line 152: 😁

lillialexis (approved):
I ended up checking out the whole repo (which I totes should have done in the first place).

I have to say, either you have spell-check in your IDE or you're a great speller! I have spell-check in my IDE, so I see _everrrrryone's_ spelling mistakes!

This is all looking spectacular!

packages/wonder-stuff-core/src/__tests__/kind-error.test.js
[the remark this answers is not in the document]
lillialexis: The webapp code could have originally been like this, but then I changed it on mobile because I noticed that Sentry was printing the error-kind twice.

packages/wonder-stuff-sentry/src/sentry-data-reducer.js
lillialexis: What else can we add to this? Jeff's programming style is very:
- defensive
- clear
- well-architected and encapsulated
- polished and well documented
- consistent with itself (← sooooo underrated!)
- easy to understand
- thoroughly tested

It's like all the #programminggoals!

packages/wonder-stuff-sentry/src/kind-sentry-error.js
somewhatabstract: There are many cases where we have data we want to log to the console or places but DO NOT want to send to sentry, so in webapp we have this division between those concepts. That's why this is done here too. Some consumers just want all the data, some, like our sentry integration, only want sentry. This achieves that in a controlled manner.

packages/wonder-stuff-sentry/src/kind-sentry-error.js
somewhatabstract: I kept them separate for now because it's not clear what options will get shared versus replaced, etc. I expect some later PR will do some consolidation as needed.

packages/wonder-stuff-core/src/__tests__/kind-error.test.js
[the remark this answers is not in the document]
lillialexis: I'm also not super attached to this either, if it's easier for you to leave as-is!

flow-typed/npm/jest-when.js
```diff
@@ -0,0 +1,28 @@
...
+
+declare module "jest-when" {
+    declare type WhenMock<TArguments: $ReadOnlyArray<any>, TReturn> = {
```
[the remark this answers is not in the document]
somewhatabstract: The `when` function provides it as typed below, based on the function or mock that is passed to it.

packages/wonder-stuff-sentry/src/kind-sentry-error.js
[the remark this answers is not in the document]
somewhatabstract: No, it doesn't because `{}` is seen as inexact but `$Partial<Options>` is exact. However, `$Shape<Options>` works. :)

packages/wonder-stuff-core/src/__tests__/kind-error.test.js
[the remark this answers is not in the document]
somewhatabstract: I think it needs changing, like you suggest - I'm going to do it in an upcoming PR rather than this one.

PR #50, opened 2021-11-11
[fei4142.1] A few fixes based on usage

somewhatabstract:
## Summary:
This makes some tweaks to our Wonder Stuff APIs based on observations made when integrating it into webapp.

1. Fixes a bug in the `browser` property of package.json files so that the mapped paths are legitimate (updated build to properly assert this too)
1. Made composite stacks an optional feature (off by default).
1. Added return types for the sentry APIs to support common Sentry API definitions
1. Less read-only types so that callers don't have to work out how to make that work nicely
1. We export `clone` per an earlier review comment
1. Added `index.d.ts` to packages so that the JSDocs will show up for folks consuming our code (does not require TypeScript - this is a docs-only thing)
1. Updated test run script for packages so that folks can, to some degree, just run the tests they want with the options they want (given that this still runs from the root, eventually, saying `yarn test error`, for example, will run matching tests in any package, not just the one where the command was executed)

Issue: FEI-4142

## Test plan:
`yarn test`
`yarn flow`

### DIFF
--- a/packages/wonder-stuff-sentry/package.json
+++ b/packages/wonder-stuff-sentry/package.json
@@ -4,7 +4,7 @@
     },
     "name": "@khanacademy/wonder-stuff-sentry",
     "version": "0.1.1",
-    "description": "",
+    "description": "Sentry integration support",
     "module": "dist/es/index.js",
     "main": "dist/index.js",
     "source": "src/index.js",
--- a/packages/wonder-stuff-sentry/package.json
+++ b/packages/wonder-stuff-sentry/package.json
@@ -9,7 +9,7 @@
     "main": "dist/index.js",
     "source": "src/index.js",
     "scripts": {
-        "test": "bash -c 'yarn --silent --cwd \"../..\" test $PWD'"
+        "test": "bash -c 'yarn --silent --cwd \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
     },
     "dependencies": {
         "@khanacademy/wonder-stuff-core": "^0.1.1"
--- a/packages/wonder-stuff-sentry/package.json
+++ b/packages/wonder-stuff-sentry/package.json
@@ -18,8 +18,8 @@
         "ws-dev-build-settings": "^0.0.3"
     },
     "browser": {
-        "dist/es/index.js": "dist/es/index.browser.js",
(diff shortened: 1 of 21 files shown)


packages/wonder-stuff-sentry/package.json
```diff
@@ -4,12 +4,12 @@
...
     "name": "@khanacademy/wonder-stuff-sentry",
     "version": "0.1.1",
-    "description": "",
+    "description": "Sentry integration support",
     "module": "dist/es/index.js",
     "main": "dist/index.js",
     "source": "src/index.js",
     "scripts": {
-        "test": "bash -c 'yarn --silent --cwd \"../..\" test $PWD'"
+        "test": "bash -c 'yarn --silent --cwd \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
```
jeresig, line 12: woah, this is complicated, what is it doing?

jeresig (approved):

packages/wonder-stuff-sentry/package.json
somewhatabstract: It is running the test script from the root `package.json`, but also allowing folks to specify test files to run by looking at the last arg and determining if it is not an option (i.e. doesn't start with `-`) and is not the `bash` command.

Really wish we could add comments to `package.json` files. :(

somewhatabstract: To break it down, when in the `packages/wonder-stuff-sentry` folder and you run `yarn test`.

`@:0` = All the args that were after the `yarn test`. e.g. for `yarn test --coverage collate-sentry-data` this would be `--coverage collate-sentry-data`

`$([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]])` = Does the last arg start with `-` or is it `bash` - this will give 0 or 1. This condenses down in the `$($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)` above to be either `$(0 && echo $PWD)` or `$(1 && echo $PWD)`. So, when the last arg starts with `-` or is `bash` we use the current working directory as our final arg, which runs all tests in our current package by generating a `yarn test` call against the root directory, passing the path of the package directory to test. If the last arg is not an option with `-` and is not `bash`, we don't add the package directory, deferring to whatever the entered last arg is, which allows things like `yarn test collate-sentry-data` to target specific tests.

PR #53, opened 2021-11-15
[fei4142.3] Add option to stringify nested context object and array values

somewhatabstract:
## Summary:
Sentry (at least on the plan we have) doesn't expand objects and arrays nested more than one level deep in contexts. This allows us to stringify these things so that they are still readable in the Sentry UX.

### Before:
image

### After:
image

Issue: FEI-4142

## Test plan:
`yarn test`
`yarn flow`
Integrate into webapp and demo it working.

### DIFF
--- a/packages/wonder-stuff-sentry/src/stringify-nested-contexts.js
+++ b/packages/wonder-stuff-sentry/src/stringify-nested-contexts.js
@@ -1,0 +1,38 @@
+// @flow
+import {safeStringify} from "@khanacademy/wonder-stuff-core";
+import type {SentryContexts} from "./types.js";
+
+export const stringifyNestedContexts = (
+    contexts: SentryContexts,
+): SentryContexts => {
+    const stringifiedContexts = {};
+
+    // For each context,
+    for (const [key, value] of Object.entries(contexts)) {
+        stringifiedContexts[key] = {};
+
+        // For each top-level property of a context (each context is an object).
+        // $FlowIgnore[incompatible-call]
+        for (const [subKey, subValue] of Object.entries(value)) {
+            if (typeof subValue === "object") {
+                if (Array.isArray(subValue)) {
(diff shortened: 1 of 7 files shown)


packages/wonder-stuff-sentry/src/stringify-nested-contexts.js
```diff
@@ -0,0 +1,38 @@
...
+    for (const [key, value] of Object.entries(contexts)) {
+        stringifiedContexts[key] = {};
+
+        // For each top-level property of a context (each context is an object).
+        // $FlowIgnore[incompatible-call]
+        for (const [subKey, subValue] of Object.entries(value)) {
+            if (typeof subValue === "object") {
+                if (Array.isArray(subValue)) {
+                    stringifiedContexts[key][subKey] = subValue.map((v) =>
+                        typeof v === "object" ? safeStringify(v) : v,
```
jeresig, line 20: I'm sure you're already aware, but just checking that you want to be handling `null` in cases like this (as `typeof null === "object"` is `true`).

jeresig (approved):
It looks like this increases the size by 144B, which isn't too bad - I think it's worth configuring our byte size action to only look at `dist/es/index.browser.js`, since that's the one where bytes are most important!

packages/wonder-stuff-sentry/src/stringify-nested-contexts.js
somewhatabstract: Thanks for checking. I wasn't sure what I wanted to do with `null`, to be honest. Perhaps I should be more specific to make it clear that we really mean it.

somewhatabstract:
> It looks like this increases the size by 144B, which isn't too bad - I think it's worth configuring our byte size action to only look at `dist/es/index.browser.js`, since that's the one where bytes are most important!

Hmm, that does seem like a worthwhile idea. I'll take a look at that.

PR #55, opened 2021-11-16
[fei4142.4] Support data validation and normalization

somewhatabstract:
## Summary:
This does two things:
1. It adds validation and normalization of Sentry data so that tag values are truncated meaningfully, and we avoid tag name clashes.
2. It removes most of the work done in the previous PR in favour of using Sentry's built-in `normalizeDepth` option.

Issue: FEI-4142

## Test plan:
`yarn test`

Integrate into webapp, updating the `normalizeDepth` option, and very that it works with extra long tag names.


jeresig (approved):
Seems ok! A couple tweaks inline - I'm worried about the increase in file size here, it's pretty substantial!

somewhatabstract:
> Seems ok! A couple tweaks inline - I'm worried about the increase in file size here, it's pretty substantial!

Agreed. I have a plan for that. I'm going to split this into dev vs prod builds. We only need the error checking in dev. So we can optionally include it. Keeping that to a separate PR.

PR #57, opened 2021-11-18
[fei4142.5] Omit parts of validation when running in production

somewhatabstract:
## Summary:
We assume that consumers will have performed due diligence and are ok with the consequences of not doing explicit error checking.

Issue: FEI-4142

## Test plan:
`yarn test`
`yarn build:prodsizecheck`


jeresig (approved):
Very nice - this seems like a big improvement, thank you!

PR #70, opened 2021-12-02
[flow-enums] Add enums support

somewhatabstract:
## Summary:
Updates some dependencies and turns on flow enum support, though we're not using it yet.

Issue: FEI-4138

## Test plan:
`yarn flow`
`yarn lint`
`yarn test`
`yarn build`


jeresig (approved):

jeresig:
Oh - wanted to ask - what are we doing to make sure that the flow enum runtime isn't being bundled into these libraries, causing us to ship two (or more) copies of them?

somewhatabstract:
> Oh - wanted to ask - what are we doing to make sure that the flow enum runtime isn't being bundled into these libraries, causing us to ship two (or more) copies of them?

At the moment, nothing. It's a super small dependency right now that I didn't do that effort as part of these PRs.

PR #86, opened 2021-12-15
[wstesting] Add wonder-stuff-testing and isolateModules helper

somewhatabstract:
## Summary:
There are likely more things to go in this package moving forward, but for now, this helper

Issue: XXX-XXXX

## Test plan:
`yarn test`

### DIFF
--- a/packages/wonder-stuff-testing/src/jest/__tests__/isolate-modules.test.js
+++ b/packages/wonder-stuff-testing/src/jest/__tests__/isolate-modules.test.js
@@ -1,0 +1,47 @@
+// @flow
+import {isolateModules} from "../isolate-modules.js";
+
+describe("#isolateModules", () => {
+    it("should execute the given function", () => {
+        // Arrange
+        const action = jest.fn();
+
+        // Act
+        isolateModules(action);
+
+        // Assert
+        expect(action).toHaveBeenCalled();
+    });
+
+    it("should return the result of the given function", () => {
+        // Arrange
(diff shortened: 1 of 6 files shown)


packages/wonder-stuff-testing/src/jest/isolate-modules.js
```diff
@@ -0,0 +1,31 @@
...
+ * This is a helper for the `jest.isolateModules` API, allowing
+ * code to avoid the clunky closure syntax in their tests.
+ *
+ * @param {() => T} action The action that contains the isolated module imports.
+ * We do it this way so that any `require` calls are relative to the calling
+ * code and not this function. Note that we don't support promises here to
+ * discourage dynamic `import` use, which doesn't play well with standard
+ * jest yet.
+ */
+export const isolateModules = <T>(action: () => T): T => {
```
somewhatabstract, line 14: This code is taken from the wonder-blocks-testing version, which this will ultimately replace.

packages/wonder-stuff-testing/src/jest/__tests__/isolate-modules.test.js
```diff
@@ -0,0 +1,47 @@
+// @flow
+import {isolateModules} from "../isolate-modules.js";
+
+describe("#isolateModules", () => {
```
somewhatabstract, line 4: I added some tests - just because we can't mock everything doesn't mean we can't test it. :)

jeresig (approved):

PR #128, opened 2022-01-25
[🔥AUDIT🔥] [againandagain] Get things closer to what Wonder Blocks does and see if it works

somewhatabstract:
🖍 _This is an audit!_ 🖍

## Summary:
So, we've dropped some of our special actions in favour of just calling package.json scripts.
Also, the coverage collection is no longer a parallel job but runs after linting/flow.

This will mean the branch protection rules need updating as I changed the name of a job.

Issue: XXX-XXXX

## Test plan:
See that #120 passes, please! Please?

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -105,6 +105,7 @@ jobs:
     - name: Run Jest with coverage
       run: yarn coverage
     - name: Upload Coverage
+      # We don't bother collecting a record of coverage for dependabot changes
       if: ${{ github.actor != 'dependabot[bot]' }}
       uses: codecov/codecov-action@v2
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/node-ci.yml
```diff
@@ -147,57 +157,29 @@ jobs:
...
-        # We don't give the GITHUB_TOKEN as that causes some odd permissions issues
-    - name: Upload Coverage
-      if: ${{ github.actor != 'dependabot[bot]' }}
-      uses: codecov/codecov-action@v2
-      with:
-        token: ${{ secrets.CODECOV_TOKEN }}
-        files: ./coverage/coverage-final.json
-        fail_ci_if_error: true
+    # Run tests for our target matrix
+    - name: Run ESLint (all files)
```
somewhatabstract, line 172: Oh no....copy pasta

.github/workflows/node-ci.yml
```diff
@@ -101,6 +101,17 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+    # Collect and record coverage
+    - name: Run Jest with coverage
+      run: yarn coverage
+    - name: Upload Coverage
+      if: ${{ github.actor != 'dependabot[bot]' }}
```
somewhatabstract, line 108: Oh, we don't want to drop this.

PR #130, opened 2022-01-25
[wait] Add wait() to WS testing

somewhatabstract:
## Summary:
This copies the `wait.js` file from webapp so we can start sharing our implementation across projects, but refactors it to get better test coverage and clarity.

Issue: XXX-XXXX

## Test plan:
`yarn test`

### DIFF
--- a/packages/wonder-stuff-testing/src/jest/wait.js
+++ b/packages/wonder-stuff-testing/src/jest/wait.js
@@ -1,0 +1,97 @@
+// @flow
+import {KindError} from "@khanacademy/wonder-stuff-core";
+import {Errors} from "../errors.js";
+
+if (typeof jest === "undefined") {
+    /**
+     * This file is for use in tests only.
+     */
+    throw new KindError(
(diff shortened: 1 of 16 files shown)


jeresig (approved):

packages/wonder-stuff-testing/src/jest/wait.js
```diff
@@ -0,0 +1,46 @@
+// @flow
+import {verifyRealTimers} from "./internal/verify-real-timers.js";
+import {unverifiedWait} from "./internal/unverified-wait.js";
+import {assertJest} from "./internal/assert-jest.js";
+
+assertJest();
```
somewhatabstract, line 6: Hmm, I wonder if after the rollup, this will do what I expect. I'm guessing not. Should probably move this to inside the various calls so it's asserted at the latest point necessary.

somewhatabstract: OK, I've modified this.

PR #169, opened 2022-03-03
[fei4349.1.addtestsfornode16] Add 16.x to workflow matrices

somewhatabstract:
## Summary:
This adds Node 16.x to the workflows and removes 14.x.

We will then want to make sure we target the lerna build to the correct version (which would be the version installed locally - currently 12.x).

Issue: FEI-4349

## Test plan:
See what happens when GitHub runs the workflows.

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node-version: [14.x]
+        node-version: [16.x]
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/node-ci.yml
```diff
@@ -137,11 +137,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest]
-        node-version: [12.x, 14.x]
+        # We want to include all the versions we support here.
+        node-version: [12.x, 16.x]
```
somewhatabstract, line 141: I kept 12.x for now - just for comparison.

jeresig (approved):

PR #170, opened 2022-03-03
[fei4220.1.changesets] Setup changeset publishing and force Node 12 or above

somewhatabstract:
## Summary:
This sets up changesets on Wonder Stuff and also sets a minimum engine of Node 12 on the root package.json.

Addresses FEI-4220 and FEI-4349

Issue: FEI-4220

## Test plan:
1. Try running `yarn` in a Node version less than 12 and see the given error
2. Try running `yarn` in a Node version greater than or equal to 12 and see things work
3. Putting this PR up should see the changeset bot comment on our changeset
4. Merging this PR should cause a Version Packages PR to get made
5. Merging the Version Packages PR should cause a new NPM release of all packages as a major version and post update notices to #frontend-infra-mobile and #frontend-infra-web


somewhatabstract:
Not quite ready yet. Need to update more actions and add SSH_PRIVATE_KEY

.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+        uses: actions/checkout@master
+        with:
+          # This makes Actions fetch all Git history so that Changesets can
+          # generate changelogs with the correct commits
+          fetch-depth: 0
+
+      - name: Setup Node.js 16.x
+        uses: actions/setup-node@master
+        with:
+          node-version: 16.x
```
somewhatabstract, line 44: Ultimately, this is what indicates what version we target. If we build under 16, we target 16 as our babel/rollup config just targets whatever version of Node it runs under. If we wanted to target 12, we'd need to change this (and probably add back in testing to verify it works on 12).

package.json
```diff
@@ -10,12 +10,16 @@
     "packages/*",
     "build-settings"
   ],
+  "engines": {
+    "node": ">=16"
+  },
```
jeresig, line 15: I don't think we should include this - this will only cause things to break for folks who aren't upgraded yet with no upside. I think it's ok to not specify the engine!

.github/workflows/release.yml
jeresig: I don't see a reason why we should break things for older versions - we just need to make sure that things work ok on newer versions! So I think it's fine to have testing for 12.x + 16.x, but only do the release for 12.x.

jeresig (changes requested):
Very cool! Really glad you're adding in Changesets. Per your note, I am going to push back. In the past when we've had strict upgrades (disallowing old versions) it's only caused us problems. I think we can make a hard push for Node v16 once everyone is upgraded, but in the meantime it seems fine to just ensure that it works in Node v16, but continue building for Node v12.

package.json
somewhatabstract: I'll change it to 12. This only affects folks developing and building Wonder Stuff, so I think it's still worth having to prevent working in the wrong version entirely.

.github/workflows/release.yml
somewhatabstract: Yeah, OK. Though I really wanted some of those new Node features :D

jeresig (approved):
Thank you!

PR #216, opened 2022-04-13
Add wonder-stuff-i18n package.

jeresig:
## Summary:
This moves the i18n logic out of Webapp and into Wonder Stuff so that it can be more-easily shared.

In the process I converted the code to use ES6 imports/exports and also use Flow types.

I had to add some logic to our build-settings so that we can generate bin scripts (since this is the first package to have them).

Issue: FEI-4434

## Test plan:
I confirmed the bin scripts worked by making a simple test file and then running `yarn build` in the root and then the following in `packages/wonder-blocks-i18n`:

```
packages/wonder-stuff-i18n (i18n-pkg) > ./dist/bin/gen-potfile.js foo.js
msgid ""
msgstr ""

#: foo.js:1
msgid "Test String"
msgstr ""

#: foo.js:3
#, python-format
msgid "Singular %(num)s"
msgid_plural "Plural %(num)s"
msgstr[0] ""
msgstr[1] ""

packages/wonder-stuff-i18n (i18n-pkg) > ./dist/bin/all-i18n-strings.js foo.js
Test String
Singular %(num)s (plural: Plural %(num)s)
```


jeresig:
I'll work on improving the test coverage before this lands! Does anyone know why the google-server/server/testing packages all are getting major version bumps? That doesn't feel accurate to me!

jeresig:
@somewhatabstract Hmm, there appears to be something wrong with the Github Action flow here, the first time it ran, it was fine (I added a new dependency, it seems to have installed it correctly). But then the second time it seems to think that the dependency no longer exists, after restoring from cache. I'm wondering if it may have cached something incorrectly. Is this the first time we've added a new build dependency, I wonder?

somewhatabstract:
> I'm wondering if it may have cached something incorrectly. Is this the first time we've added a new build dependency, I wonder?

That seems like the most obvious answer. I've definitely done things with new deps before. Is the cache key correct to detect changes? Perhaps it's hashing the wrong files?

somewhatabstract:
Seems like you didn't add the updated `yarn.lock` to your PR.

packages/wonder-stuff-i18n/src/bin/gen-potfile.js
```diff
@@ -0,0 +1,29 @@
...
+ * You can specify files on the command-line and this script will only extract
+ * strings from those files, instead of from all JavaScript file.
+ *
+ * USAGE:
+ *
+ *   # Extract strings from all JS files and export to a shared POT file
+ *   ./gen-potfile.js javascript/**\/*.{js,jsx} > JavaScript.pot
+ *
+ *   # Extract strings from a specific file and output to STDOUT
+ *   ./gen-potfile.js javascript/about/about.jsx
```
kevinbarabash, line 17: nit: update usage to show how this would be used when installed as an npm package.

packages/wonder-stuff-i18n/src/bin/gen-potfile.js
```diff
@@ -0,0 +1,29 @@
...
+import {
+    getFilesToExtractFrom,
+    getPOTFileStringFromFiles,
+} from "../utils/pofile-utils.js";
+
+// Get the files that we want to process.
+const files = getFilesToExtractFrom(process.argv.slice(2));
+
+// Generate the POT file as a string and output it to STDOUT
+// eslint-disable-next-line no-console
```
kevinbarabash, line 28: style nit: consider adding an override for wonder-stuff-i18n/src/bin to have this not be reported.

packages/wonder-stuff-i18n/src/index.js
```diff
@@ -0,0 +1,13 @@
...
+export {
+    getFilesToExtractFrom,
+    getPOTFileStringFromFiles,
+} from "./utils/pofile-utils.js";
+export {
+    getIgnoreGlobs,
+    getI18nStringsFromString,
+    translateString,
+} from "./utils/i18n-utils.js";
+export {getEmojiForLocale} from "./utils/emoji-for-locale.js";
```
kevinbarabash, line 13: 😍

packages/wonder-stuff-i18n/src/plugins/i18n-plugin.js
```diff
@@ -0,0 +1,410 @@
...
+        start: (string) => void,
+        end: (string) => void,
+        ...
+    },
+    getI18nStrings: (locale: string) => Promise<TranslatedLocaleStrings>,
+    getLocalePath: (locale: string) => string,
+    shouldLocalizeAsset: (assetName: string) => boolean,
+    shouldProcessAsset: (assetName: string) => boolean,
+    shouldLocalizeManifest: (assetName: string, locale: string) => boolean,
+};
```
kevinbarabash, line 88: I think it would be easier to see the differences (and maintain the consistencies) between these if `IncomingOptions` was defined in terms of `Options`, e.g.
```
type IncomingOptions = {
   ...Options,
   getLocalePath?: Options['getLocalePath'],
   shouldLocalizeAsset?: Options['shouldLocalizeAsset'],
   shouldProcessAsset?: Options['shouldProcessAsset'],
   shouldLocalizeManifest?: Options['shouldLocalizeManifest'],
};
```

packages/wonder-stuff-i18n/src/plugins/i18n-plugin.js
```diff
@@ -0,0 +1,410 @@
...
+            this.onEmit(options, callback),
+        );
+    }
+
+    /**
+     * Log out something to the console, if we're not silent.
+     *
+     * @param  {...any} args the args to log out to the console
+     */
+    log(...args: Array<$FlowFixMe>) {
```
kevinbarabash, line 141: I think using `Array<any>` is appropriate here.  This is how flow itself types `console.log()`, see https://github.com/facebook/flow/blob/main/lib/core.js#L2453.  Using `$FlowFixMe` may tempt some people to figure out a fix even though there isn't one.

packages/wonder-stuff-i18n/src/plugins/i18n-plugin.js
```diff
@@ -0,0 +1,410 @@
...
+        assets,
+        translatedStrings,
+        localizeRuntimeFiles,
+        hashMap,
+    }: {
+        assets: Assets,
+        translatedStrings: TranslatedStrings,
+        localizeRuntimeFiles: boolean,
+        hashMap?: HashMaps,
+    }): HashMaps {
```
kevinbarabash, line 202: Seeing all of these flow types in a webpack plugin fills my heart with immense joy. 😍

packages/wonder-stuff-i18n/src/plugins/i18n-plugin.js
```diff
@@ -0,0 +1,410 @@
...
+        return hashMap;
+    }
+
+    /**
+     * Update manifest files so that they point to the correct location of
+     * the translated files.
+     *
+     * @param {assets: Object<string, Object>, hashMap: <string, string>} options
+     */
+    localizeManifests({assets, hashMap}: {assets: Assets, hashMap: HashMaps}) {
```
kevinbarabash, line 327: TypeScript supports inlay hints which removes the need for this style of args (except when there's a bunch of optional args in which case an "options" param makes sense).  Having these be separate params would cut down on some of the visualize noise of having to duplicate the field names.  See https://code.visualstudio.com/docs/languages/typescript#_inlay-hints for details about inlay hints (VSCode has a bunch of options to configure in which situations they show up).  There's also a keyboard shortcut to toggle them which is nice.

packages/wonder-stuff-i18n/src/plugins/i18n-plugin.js
```diff
@@ -0,0 +1,410 @@
...
+    /**
+     * Logic to run when emitting the files. This will get the translated
+     * strings, translate the assets, and update the manifest files.
+     *
+     * @param {Object} options compilation options from Webpack
+     * @param {Function} callback callback function to call when done
+     * @returns void
+     */
+    async onEmit(
+        {assets}: {assets: Assets},
```
kevinbarabash, line 374: I wish there was a webpack class that plugins could extend so that we could get some of these types for free.  No action req'd.

packages/wonder-stuff-i18n/src/utils/emoji-for-locale.js
```diff
@@ -0,0 +1,48 @@
...
+    accents: "✨", // Accents
+    bg: "🇧🇬", // Bulgarian --Bulgaria
+    bn: "🇮🇳", // Bengali -- India
+    boxes: "📦", // Unicode boxes
+    cs: "🇨🇿", // Czech -- Czech Republic
+    da: "🇩🇰", // Danish -- Denmark
+    de: "🇩🇪", // German -- Germany
+    en: "🇬🇧", // English -- UK
+    "en-pt": "☠️", // Pirate english
+    es: "🇲🇽", // Spanish
```
kevinbarabash, line 17: I always get a kick out of this choice of flag.

packages/wonder-stuff-i18n/src/utils/emoji-for-locale.js
```diff
@@ -0,0 +1,48 @@
...
+    "zh-hans": "🇹🇼", // Mandarin -- Taiwan
+};
+
+/**
+ * Given a locale, get an emoji representing that locale.
+ *
+ * @param {string} locale the locale string to get the emoji for
+ */
+export const getEmojiForLocale: (locale: string) => string = (locale: string) =>
+    LOCALE_TO_EMOJI_MAP[locale] || "";
```
kevinbarabash, line 48: This amount of typing to export a function makes me sad.  TypeScript will be much better in this regard.

packages/wonder-stuff-i18n/src/utils/emoji-for-locale.test.js
```diff
@@ -0,0 +1,37 @@
...
+    it("handles an unknown locale", () => {
+        // Arrange
+        const locale = "asdf";
+
+        // Act
+        const result = getEmojiForLocale(locale);
+
+        // Assert
+        expect(result).toEqual("");
+    });
```
kevinbarabash, line 36: When we filter out `en` out of the list of locales to translate I wonder if we should also filter out unknown locales as well.

packages/wonder-stuff-i18n/src/utils/extract-i18n.js
```diff
@@ -0,0 +1,508 @@
...
+ *  msgids: Array<string>,
+ *  // The comments associated with the string.
+ *  comments: Array<string>,
+ *
+ * @param {string} textString
+ * @returns
+ */
+export function* extractStrings(
+    textString: string,
+): Generator<ExtractedString, void, void> {
```
kevinbarabash, line 414: That's cool that you can specify the type of arg to `next()`.  It's too bad the types don't default to `void` and `void` for the last two args.

packages/wonder-stuff-i18n/src/utils/i18n-utils.js
```diff
@@ -0,0 +1,113 @@
...
+    const strings = [];
+
+    // Get all the strings that need to be replaced, but in reverse
+    // order, so that it's easier to insert them in the right place.
+    for (const extractedString of extractStrings(textString)) {
+        strings.unshift(extractedString);
+    }
+
+    return strings;
+};
```
kevinbarabash, line 63: Now that we have types for things I feel much better having a bunch of different things using `strings` as their identifier. 🙂

packages/wonder-stuff-i18n/src/utils/i18n-utils.js
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * Insert translated strings into a code string, returning the fully translated
+ * code string.
+ *
+ * @param {string} textString the string to translate, containing code.
+ * @param {Object<string, string|Object>} translatedStrings the translated strings
+ * to insert into code string.
+ * @param {Array<Object>} strings the strings that've been pre-extracted from the
```
kevinbarabash, line 72: super style nit: you could use `TranslatedLocaleStrings` and `Array<ExtractedString>` here to match the flow types.

packages/wonder-stuff-i18n/src/utils/localize-file.js
```diff
@@ -0,0 +1,143 @@
...
+        if (oldHash !== newHash) {
+            hashMap[oldHash] = newHash;
+
+            // We also need to update any existing mappings that already are
+            // in place to point to the new hash as well
+            for (const hash of Object.keys(hashMap)) {
+                if (hashMap[hash] === oldHash) {
+                    hashMap[hash] = newHash;
+                }
+            }
```
kevinbarabash, line 138: I'm not sure what this security check is actually testing:
```
> const a = {};
undefined
> a
{}
> a[proto] = {foo: () => console.log("foo")}
Uncaught ReferenceError: proto is not defined
> 
```

packages/wonder-stuff-i18n/src/utils/pofile-utils.js
```diff
@@ -0,0 +1,170 @@
...
+
+import type {ExtractedString} from "./extract-i18n.js";
+
+type POItem = {|
+    msgid: string,
+    msgid_plural: string,
+    flags: {[string]: boolean},
+    references: Array<string>,
+    extractedComments: Array<string>,
+|};
```
kevinbarabash, line 19: If you wanted to be fancy you could create a libdef for `pofile` in flow-typed.  What you have it totally fine as well.  No action req'd.

packages/wonder-stuff-i18n/src/utils/pofile-utils.js
```diff
@@ -0,0 +1,170 @@
...
+    poItem.references[0] += ` ${fileName}:${linePos}`;
+
+    // Comments are added seemingly in reverse order. We maintain that
+    // here to match the legacy PO file output.
+    for (const comment of comments) {
+        if (!poItem.extractedComments.includes(comment)) {
+            poItem.extractedComments.unshift(comment);
+        }
+    }
+};
```
kevinbarabash, line 101: TIL

kevinbarabash (approved):
LGTM.  Since the files were moved from webapp and already reviewed there I focused on the types which is the new part.  There are few nits, mostly already usage comments/messages that should be updated.  I'm really happy that all of this code is now well typed in addition to be well tested.  🎉 🤘

packages/wonder-stuff-i18n/src/bin/all-i18n-strings.js
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env node
+// @flow
+/* istanbul ignore file */
+/* eslint-disable no-console */
+/**
+ * Extract all i18n strings from JS files.
+ *
+ * USAGE:
+ *
+ *   ./webpack/tools/all-i18n-strings.js
```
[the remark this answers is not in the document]
jeresig: Updated!

packages/wonder-stuff-i18n/src/bin/find-new-i18n-strings.js
```diff
@@ -0,0 +1,51 @@
...
+    console.log(
+        "Your pull request has introduced new strings that need to " +
+            `be translated. Please [submit a Translation Request](${link}) ` +
+            "with the following strings so that the translation team can get started!\n",
+    );
+    console.log("### New Strings:\n");
+    for (const msgid of afterStrings) {
+        console.log(` * \`${msgid}\``);
+    }
+    console.log(`### » 🎉 [Submit Translation Request](${link}) 🎉 «`);
```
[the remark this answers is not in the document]
jeresig: They don't - the link auto-populates the description with the strings in it. Hopefully saves them some time!

packages/wonder-stuff-i18n/src/bin/gen-potfile.js
jeresig: Done!

packages/wonder-stuff-i18n/src/bin/gen-potfile.js
jeresig: Good call, fixed!

packages/wonder-stuff-i18n/src/plugins/i18n-plugin.js
jeresig: Hmm... so we have a rule in this repo that prevents all usage of `any`, which is why I had to change this! Is there a way around this?

packages/wonder-stuff-i18n/src/plugins/i18n-plugin.js
jeresig: Hmm... so I understand what you're suggesting, but I really want to have the docs be on the true "Options" object and have the "InternalOptions" be the one that's different (and undocumented). I'll upload what I have, but I don't think using `Options['getLocalePath']` will work in that case as it'll copy over the optional-ness of the property as well.

packages/wonder-stuff-i18n/src/plugins/i18n-plugin.js
jeresig: Yeah, that would be nice!

packages/wonder-stuff-i18n/src/utils/emoji-for-locale.js
jeresig: Oh - I think I messed this up, actually - I just changed it to use `: string` for the return and it works now.

packages/wonder-stuff-i18n/src/utils/emoji-for-locale.test.js
jeresig: Hmm - so right now we don't really have a concept of "known" and "unknown" locales (only in-so-much as when we don't get any strings for a locale). So I'm not entirely sure it makes sense to do that (I'd rather just error out because we can't find the strings).

packages/wonder-stuff-i18n/src/utils/localize-file.js
jeresig: Hmm? I'm not sure I follow... we're inserting a property whose value is stored in hash.

jeresig:
@somewhatabstract I figured out the problem! Although I'm unsure how to resolve it off-hand. So when the yarn.lock changes we build the dependencies from scratch, however, we *also* check out the previous commit and then run `yarn install` again, which causes us to cache the *old* `node_modules` rather than the new one.

PR #277, opened 2022-06-07
Fix invalid yaml 🤦‍♂️ in the node-ci workflow

jaredly:
## Summary:
ugh
I wonder if we can set up prettier or eslint for our .yml files? That would at least make these errors more obvious.

Issue: XXX-XXXX

## Test plan:
🤞

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -112,10 +112,10 @@ jobs:
         # Build production
         build-script: "build:prodsizecheck"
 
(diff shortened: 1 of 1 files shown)


somewhatabstract (approved):
Ah, thanks for fixing this. And yes, a yaml linter would be nice.

somewhatabstract:
@jaredly  Looks like the jest action doesn't like having no files.

jaredly:
oh it doesn't like it when I reference a nonexistent step 😭 fixed 🤞

PR #278, opened 2022-06-07
Move eslint-config-khan into wonder-stuff

kevinbarabash:
## Summary:
This will make it easier to publish updates moving forward.

In addition to moving the code, I've also update the config to use eslint-plugin-flowtype instead of eslint-plugin-ft-flow.  This will all other repos that depend on eslint-config-khan (such as wonder-blocks) to be updated to use eslint-plugin-ft-flow.  The new eslint plugin resolves an issue involving Flow enums, see FEI-4172 for more details.

Issue: FEI-4172

## Test plan:
- n/a, I need to publish before I can test this because eslint has issues resolving things when using yarn link


build-settings/rollup.config.js
```diff
@@ -267,7 +267,9 @@ const getPkgNames = (commandLineArgs) => {
     const {configPackages} = commandLineArgs;
 
     // Get the list of packages that we have in our packages folder.
-    const actualPackages = fs.readdirSync("packages");
+    const actualPackages = fs
+        .readdirSync("packages")
+        .filter((name) => name.startsWith("wonder-stuff-"));
```
kevinbarabash, line 272: There's no build step for `eslint-config-khan`.  I'll add a comment in the code about this.

.eslintrc.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable import/no-commonjs */
 module.exports = {
-    extends: ["@khanacademy"],
+    extends: ["./packages/eslint-config-khan/index.js"],
```
kevinbarabash, line 3: The benefit of using this here is that running `yarn lint` will run any changes to the config against the code in this repo which gives us some information about how the config is performing before we actually publish changes to it.

packages/eslint-config-khan/package.json
```diff
@@ -0,0 +1,27 @@
...
+        "eslint-plugin-ft-flow": "^5.7.2",
+        "eslint-plugin-graphql": "^4.0.0",
+        "eslint-plugin-import": "^2.23.4",
+        "eslint-plugin-jsx-a11y": "^6.4.1",
+        "eslint-plugin-prettier": "^3.4.0",
+        "eslint-plugin-react": "^7.24.0",
+        "eslint-plugin-react-hooks": "^4.2.0",
+        "eslint-plugin-react-native": "^3.11.0",
+        "eslint-plugin-react-native-animation-linter": "^0.1.2"
+    }
```
kevinbarabash, line 26: We might actually want to make these real dependencies in the future instead of peer deps.  I'm going to punt on that for now.

packages/wonder-stuff-core/src/kind-error.js
```diff
@@ -111,7 +111,7 @@ export class KindError extends Error {
             stripStackFrames,
             minimumFrameCount,
             compositeStack,
-        }: Options = {},
+        }: Options = Object.freeze({}),
```
kevinbarabash, line 114: This was a new Flow error that popped up.  Not sure why it wasn't caught before.

build-settings/rollup.config.js
jeresig: It doesn't feel like this is how we want to configure this, as otherwise this is going to fail (in a very confusing way) the next time we add a non-wonder-stuff package that does have a build step. Instead, it would be better to add something to the `package.json` indicating that there is no build step, which we then read in `getPackageInfo` and filter out before we call `createConfig` in `createRollupConfig`.

packages/eslint-config-khan/.prettierrc.js
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable import/no-commonjs */
```
jeresig, line 1: Do we need this file? Do we already have a global prettierrc config for all of wonder-stuff?

jeresig (changes requested):
Thanks for moving this in! A couple changes inline which should help to make this more future-proof.

build-settings/rollup.config.js
kevinbarabash: Good idea.

packages/eslint-config-khan/.prettierrc.js
kevinbarabash: The wonder-stuff .prettierrc.js imports this.  Right now it's importing the published version.  I'll update it to import this one directly.

jeresig (approved):
Thanks for all the fixes!

kevinbarabash:
I'm going to exclude eslint-config-khan from coverage report before landing since it doesn't make sense to include.

PR #285, opened 2022-06-08
Fix pre-publish check to work with eslint-config-khan

kevinbarabash:
## Summary:
pre-publish-check-ci.js makes a lot of assumptions about what various fields should be in the package.json fiels for each package.  eslint-config-khan, doesn't conform to these because it doesn't require a build step.  I've updated the script to do a different check for fields with 'nobuild: true' in their package.json.  We check to make sure that the 'main' field exists on disk.

I've also updated node-ci.yml to run the pre-publish check during PRs.  This will help identify issues during PRs making it easier for people to correct anything that's not quite right.

NOTE: now that we're running pre-publish-check-ci.js on each PR, each PR needs to have a changesets file even if it's empty.  An empty on can be generated by running `yarn changeset add --empty`.

Issue: None

## Test plan:
- rm packages/eslint-config-khan/index.js
- node utils/pre-publish-check-ci.js, see an error

```
kevinbarabash@Kevins-MacBook-Pro wonder-stuff % rm packages/eslint-config-khan/index.js 
kevinbarabash@Kevins-MacBook-Pro wonder-stuff % node utils/pre-publish-check-ci.js     
ERROR: @khanacademy/eslint-config's "main" field is set to index.js, but no file exists at that path.
```

- git restore packages/eslint-config-khan/index.js
- node utils/pre-publish-check-ci.js, see no errors


.github/workflows/node-ci.yml
```diff
@@ -174,3 +174,27 @@ jobs:
...
+    - name: Install & cache node_modules
+      uses: Khan/actions@shared-node-cache-v0
+      with:
+        node-version: ${{ matrix.node-version }}
+
+    - name: Build so that dist files will exist
+      run: yarn build
+
+    - name: Pre-publish check
+      run: node utils/pre-publish-check-ci.js
```
somewhatabstract, line 200: Since we already do most of this work in the lint job, could you just add this command as a job in that job instead? I think it will be better that way and cheaper on github action minutes.

utils/pre-publish-check-ci.js
```diff
@@ -19,6 +20,19 @@ fg(path.join(__dirname, "..", "packages", "**", "package.json")).then(
...
+                if (!fs.existsSync(mainPath)) {
+                    const {name, main} = pkgJson;
+                    console.error(
+                        `ERROR: ${name}'s "main" field is set to ${main}, but no file exists at that path.`,
+                    );
+                    process.exit(1);
+                }
+                continue;
+            }
+
```
somewhatabstract, line 35: Could you make this a function like the other checks? Just for readability. Something like `checkNoBuildMain`?

somewhatabstract: Also, should this be in the `!checkPrivate` condition?

.changeset/tall-cows-ring.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
somewhatabstract, line 2: Is this what we have to do now that we have a changeset rule on PRs?

.github/workflows/node-ci.yml
```diff
@@ -117,6 +117,9 @@ jobs:
       with:
         exclude: .github/,.storybook/
 
+    - name: Pre-publish check
+      run: node utils/pre-publish-check-ci.js
```
somewhatabstract, line 121: can you move this and the changeset check to above the build size check? The build size check changes what code is checked out and should always go last.

somewhatabstract (changes requested):
Looks good, just one more change please. Apologies. ❤️

.changeset/tall-cows-ring.md
kevinbarabash: Yes, that's correct.

somewhatabstract: Do we have a handy command to make this easier, like `yarn nochangeset`?

.github/workflows/node-ci.yml
```diff
@@ -97,6 +97,14 @@ jobs:
         files: ./coverage/coverage-final.json
         fail_ci_if_error: true
 
+    - uses: Khan/actions@check-for-changeset-v0
+      if: github.actor != 'dependabot[bot]' && github.actor != 'dependabot-preview[bot]'
+      with:
+        exclude: .github/,.storybook/
+
+    - name: Pre-publish check
+      run: node utils/pre-publish-check-ci.js
```
somewhatabstract, line 106: Thank you. I was going to say add a comment about the size check going last, but I see there already is one. Could you add a similar comment after it saying NO MORE STEPS AFTER THIS STEP or something?

somewhatabstract (approved):
Thanks. One last minor thing if you are able, but otherwise, this is great!

somewhatabstract:
Oh, looks like there's something not quite right about the pre-publish check invocation still.

.github/workflows/node-ci.yml
kevinbarabash: I can do that. 🙂

.github/workflows/node-ci.yml
```diff
@@ -112,10 +120,9 @@ jobs:
         # Build production
         build-script: "build:prodsizecheck"
 
-    - uses: Khan/actions@check-for-changeset-v0
-      if: github.actor != 'dependabot[bot]' && github.actor != 'dependabot-preview[bot]'
-      with:
-        exclude: .github/,.storybook/
+    #
+    # Do not place any steps after "Check Builds"
+    #
```
somewhatabstract, line 125: ❤️

package.json
```diff
@@ -71,6 +71,7 @@
     "lint": "eslint --config .eslintrc.js packages",
     "lint:ci": "eslint --config .eslintrc.js",
     "publish:ci": "node utils/pre-publish-check-ci.js && git diff --stat --exit-code HEAD && yarn build && changeset publish",
-    "test": "yarn jest"
+    "test": "yarn jest",
+    "nochangeset": "yarn changeset add --empty"
```
somewhatabstract, line 75: :tada:

PR #314, opened 2022-07-08
[fei4636.1] Add values() method

somewhatabstract:
## Summary:
This adds a version of `values()` that provides strong typing of the returned array, rather than `Array<mixed>` as `Object.values()` would.

Issue: FEI-4636

## Test plan:
`yarn test`
`yarn flow`

### DIFF
--- a/.changeset/warm-terms-care.md
+++ b/.changeset/warm-terms-care.md
@@ -1,5 +1,5 @@
 ---
-"@khanacademy/wonder-stuff-core": major
+"@khanacademy/wonder-stuff-core": minor
 ---
 
 Add `values` method as strongly-typed alternative to `Object.values`
--- a/.changeset/warm-terms-care.md
+++ b/.changeset/warm-terms-care.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-stuff-core": major
+---
+
+Add `values` method as strongly-typed alternative to `Object.values`
--- a/packages/wonder-stuff-core/src/__tests__/values.flowtest.js
+++ b/packages/wonder-stuff-core/src/__tests__/values.flowtest.js
@@ -1,0 +1,36 @@
+// @flow
+import {values} from "../values.js";
+
+{
+    // should type returned array element with union of value types of passed object
+    const obj1 = {
+        a: 1,
+        b: "2",
+        c: [3, 4],
+    };
+    const obj1Values = values(obj1);
+
+    // This works because the variable is typed to all the possible value types.
+    const _: number | string | Array<number> = obj1Values[0];
+
+    // This errors because the variable is only typed to string, but the value
+    // could be a number, string, or array of numbers.
+    // $FlowExpectedError[incompatible-type]
+    const __: string = obj1Values[1];
+}
+
+{
+    // should error if passed object values do not match parameterized type
+    const obj2 = {
+        a: 1,
+        b: "2",
+    };
+
+    // This errors because the return type of values() is not Array<number>
+    // $FlowExpectedError[incompatible-type]
+    const _: Array<number> = values(obj2);
+
+    // This errors because the object does not have values that are all numbers.
+    // $FlowExpectedError[incompatible-call]
+    const __ = values<number>(obj2);
+}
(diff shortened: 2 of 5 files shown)


.changeset/warm-terms-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-stuff-core": major
```
benchristel, line 2: Is this really a major release? It should be backward-compatible, no?

packages/wonder-stuff-core/src/__tests__/values.flowtest.js
```diff
@@ -0,0 +1,36 @@
...
+    };
+
+    // This errors because the return type of values() is not Array<number>
+    // $FlowExpectedError[incompatible-type]
+    const _: Array<number> = values(obj2);
+
+    // This errors because the object does not have values that are all numbers.
+    // $FlowExpectedError[incompatible-call]
+    const __ = values<number>(obj2);
+}
```
benchristel, line 36: I'd be interested to see what type Flow infers for the array if the object has no properties. It also might be worth including a flowtest that uses an object-as-map type like `{|[string]: number|}`. I think Flow will infer struct-like types for the objects in the current tests.

packages/wonder-stuff-core/src/__tests__/values.test.js
```diff
@@ -0,0 +1,36 @@
...
+            c: [3, 4],
+        };
+        jest.spyOn(Object, "values").mockReturnValue([1, 2, 3]);
+
+        // Act
+        const result = values(obj);
+
+        // Assert
+        expect(result).toEqual([1, 2, 3]);
+    });
```
benchristel, line 35: What's the purpose of mocking `Object.values` here? IMO a non-mocked test would:

- illustrate usage and expected results better, by showing real inputs and outputs.
- avoid coupling the tests to the implementation (perhaps letting us refactor to a different implementation with test coverage, if by some chance we find that Object.values doesn't do what we want).

Perhaps, though, we want these tests to enforce the invariant that `values` is a pass-through wrapper around `Object.values`, since that's how we want to define `values`. If that's the case, I think the tests could use a comment (or maybe different test titles?) explaining the intent.

packages/wonder-stuff-core/src/values.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+/**
+ * Return an array of the enumerable properties of an object.
+ *
+ * @param {$ReadOnly<{[mixed]: V}>} obj The object for which the values are
+ * to be returned.
+ * @returns {Array<V>} An array of the enumerable properties of an object.
```
benchristel, line 7: Nit: this should probably say "enumerable properties of _the_ object", or perhaps just "enumerable property values" (since the values, not the properties, are returned).

benchristel (approved):
Left a few suggestions inline, but overall this LGTM!

packages/wonder-stuff-core/src/values.js
somewhatabstract: I based this language off the language that is documenting Object.values

packages/wonder-stuff-core/src/__tests__/values.test.js
somewhatabstract: It seems appropriate to just test that it returns what `Object.values` returns rather to retest what that should be, especially since that's our aim with these wrappers.

I think the test title explains that. It says that it should return what `Object.values` returns and it does just that. I don't think there's anything to adjust here.

packages/wonder-stuff-core/src/__tests__/values.flowtest.js
somewhatabstract: What outcome would these tests be catching/looking for? Generally these tests are to assert the types are working and to give us a signal if they break. What specifically would the test case be for in these examples?

.changeset/warm-terms-care.md
somewhatabstract: Yeah, probably not.

kevinbarabash (approved):
LGTM.  Thanks for adding this to wonder-stuff.

jeresig (approved):

packages/wonder-stuff-core/src/__tests__/values.flowtest.js
benchristel: I expect `values` to accept objects-as-maps, and infer the correct output type, so it seems valuable to have a flowtest  ensuring that works. For the empty-object case, I was mostly just curious what the return type would be. I agree it doesn't seem very useful to have a test for that.

PR #315, opened 2022-07-08
[fei4636.2] Add keys() method

somewhatabstract:
## Summary:
This adds an wrapper around `Object.keys` that returns `Array<$Keys<O>>` instead of `Array<string>`.

Issue: FEI-4636

## Test plan:
`yarn test`
`yarn flow`

### DIFF
--- a/packages/wonder-stuff-core/src/__tests__/keys.test.js
+++ b/packages/wonder-stuff-core/src/__tests__/keys.test.js
@@ -1,0 +1,36 @@
+// @flow
+import {keys} from "../keys.js";
+
+describe("#keys", () => {
+    it("should call Object.keys with the given object", () => {
+        // Arrange
+        const keysSpy = jest.spyOn(Object, "keys");
+        const obj = {
+            a: 1,
+            b: "2",
+            c: [3, 4],
+        };
+
+        // Act
+        keys(obj);
+
+        // Assert
+        expect(keysSpy).toHaveBeenCalledWith(obj);
+    });
+
+    it("should return the result of Object.keys", () => {
+        // Arrange
+        const obj = {
+            a: 1,
+            b: "2",
+            c: [3, 4],
+        };
+        jest.spyOn(Object, "keys").mockReturnValueOnce([1, 2, 3]);
+
+        // Act
+        const result = keys(obj);
+
+        // Assert
+        expect(result).toEqual([1, 2, 3]);
+    });
+});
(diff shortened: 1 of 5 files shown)


jeresig (approved):

packages/wonder-stuff-core/src/__tests__/keys.test.js
```diff
@@ -0,0 +1,36 @@
...
+    });
+
+    it("should return the result of Object.keys", () => {
+        // Arrange
+        const obj = {
+            a: 1,
+            b: "2",
+            c: [3, 4],
+        };
+        jest.spyOn(Object, "keys").mockReturnValueOnce([1, 2, 3]);
```
kevinbarabash, line 28: nit: having the mock return `[1,2,3]` instead of `["a", "b", "c"]` makes for a confusing test.

kevinbarabash (approved):
LGTM.  One minor nit in the tests, but it isn't blocking.  Thanks for adding this to wonder-stuff.

packages/wonder-stuff-core/src/__tests__/keys.test.js
benchristel: Hmm, I actually think using realistic values tends to make it less obvious what's mocked vs. not. In the interest of brevity and clarity of intent, I might prefer something even more obviously not-real:

```js
jest.spyOn(Object, "keys").mockReturnValueOnce("the return value of Object.keys");
expect(keys({})).toBe("the return value of Object.keys");
```

packages/wonder-stuff-core/src/keys.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+/**
+ * Return an array of the enumerable keys of an object.
+ *
+ * @param {$ReadOnly<{[string]: mixed}>} obj The object for which the values are
+ * to be returned.
+ * @returns {Array<$Keys<O>>} An array of the enumerable keys of an object.
+ */
+export function keys<O: {[string]: mixed}>(obj: O): Array<$Keys<O>> {
```
benchristel, line 9: Good call forcing the type of `O` to be a subtype of `{[string]: mixed}`—otherwise `keys` would not behave according to its type if `O` were typed to have non-string keys. IMO it's pretty weird that Flow lets you have types like `{[number]: string}`, since that doesn't match the runtime types.

benchristel (approved):
LGTM!

PR #320, opened 2022-07-12
[fixbuild] Update to Node 16

somewhatabstract:
## Summary:
Did this while investigating an issue and thought we should keep it since we are targetting Node 16 now.

Issue: XXX-XXXX

## Test plan:
`yarn test`
`yarn build`
Check the actions all work as expected.

### DIFF
--- a/packages/wonder-stuff-i18n/package.json
+++ b/packages/wonder-stuff-i18n/package.json
@@ -3,7 +3,7 @@
         "access": "public"
     },
     "engines": {
-        "node": ">=12"
+        "node": ">=16"
     },
     "name": "@khanacademy/wonder-stuff-i18n",
     "version": "1.1.0",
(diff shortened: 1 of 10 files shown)


packages/wonder-stuff-i18n/package.json
```diff
@@ -3,7 +3,7 @@
         "access": "public"
     },
     "engines": {
-        "node": ">=12"
+        "node": ">=16"
```
jeresig, line 6: Why do only some packages have a node 16 restriction?

jeresig (approved):

packages/wonder-stuff-i18n/package.json
somewhatabstract: I don't know the answer to that.

somewhatabstract: I can modify it so they all do

jeresig: Well, either all do or none do - whichever!

PR #321, opened 2022-07-12
[fixkeystypes] Improve keys typing

somewhatabstract:
## Summary:
I found that the typing was not helpful so I added a test case for the issue I experienced and then tweaked the function definition so that all the flow tests passed.

No functional change here, purely types.

Issue: FEI-4655

## Test plan:
`yarn flow`


jeresig (approved):

packages/wonder-stuff-core/src/keys.js
```diff
@@ -2,10 +2,10 @@
 /**
  * Return an array of the enumerable keys of an object.
  *
- * @param {$ReadOnly<{[string]: mixed}>} obj The object for which the values are
+ * @param {$ReadOnly<{[string]: V}>} obj The object for which the values are
```
benchristel, line 5: I'm curious if adding `$ReadOnly` to the type of `obj` would have fixed this issue. In any case, it seems like the doc comment should match the actual types—either the `$ReadOnly` here should be removed, or `$ReadOnly` should be added to the function signature.

somewhatabstract: Oh, good call. Let me try that

somewhatabstract: Yep, that works. Good spot, @benchristel. Thanks! :)

PR #329, opened 2022-07-14
[objasinterface] Make sure that keys/values/entries are typed to work against class instances too

somewhatabstract:
## Summary:
Noticed in webapp that some uses are passing a class instance and we should support that. Flow error indicated using an interface rather than object type so that it would match both objects and class instances. So this does that.

Issue: XXX-XXXX

## Test plan:
`yarn flow`


kevinbarabash (approved):
LGTM. Thanks for tweaking the Flow types to support class instances. 💚

PR #380, opened 2022-09-01
[fei4139.ws.1] Updating to Jest v29

somewhatabstract:
## Summary:
This updates Wonder Stuff to Jest 29. Notable changes:

- Snapshots are updated (the new snapshot format is nicer to read IMO)
- The babel config hack we had is not required, but we do need a root babel config for jest to find as the transformer API uses the working directory to locate the babel config it seems, otherwise, snapshot updates fail

Issue: FEI-4139

## Test plan:
`yarn test`

Also, edit an inline snapshot to be wrong and then use `yarn test -u` to update snapshots to see that it works.


.changeset/yellow-flies-happen.md
```diff
@@ -0,0 +1,2 @@
+---
```
somewhatabstract, line 1: **note:** There shouldn't be any actual package changes out of this work.

flow-typed/npm/jest_v28.x.x.js
```diff
@@ -1,7 +1,7 @@
-// flow-typed signature: 5ddcf688200e3506308fdcfa78ca48d9
-// flow-typed version: 644a595e77/jest_v27.x.x/flow_>=v0.134.x
+// flow-typed signature: fd6b1d81136037fad6c4e96fa7354af4
+// flow-typed version: 3153a3ba01/jest_v28.x.x/flow_>=v0.134.x
```
somewhatabstract, line 2: **note:** There aren't flow types for v29. Not sure what we can add that would be missing, but updating to v28 to get us closer to right. :)

packages/wonder-stuff-testing/src/jest/isolate-modules.js
```diff
@@ -13,10 +13,10 @@ import {assertJest} from "./internal/assert-jest.js";
  * discourage dynamic `import` use, which doesn't play well with standard
  * jest yet.
  */
-export const isolateModules = <T>(action: () => T): T => {
+export const isolateModules = <T>(action: () => ?T): ?T => {
```
jeresig, line 16: Huh, why did this change? Wouldn't this also require a minor version update for the wonder-stuff-testing package?

jeresig (approved):

packages/wonder-stuff-testing/src/jest/isolate-modules.js
somewhatabstract: I am honestly not sure. Flow started complaining about it - I can only assume some auto-update of flow missed it before.

The code already worked this way so it's only a type change; existing code shouldn't behave any differently from this so I didn't feel a minor update was worth it.

PR #428, opened 2022-10-12
Add 'dataFactoryFor' function to help generate mock data

kevinbarabash:
## Summary:
This PR extracts 'getCreateMockData' from https://github.com/Khan/webapp/pull/8976 so that we can use it other repo than just 'webapp'.  This PR also:
- renames this function to 'dataFactoryFor'
- updates the examples in the documentation so that they show valid usage
- fixes a bug in the code that could've resulting in mutation of the partial object being passed to the data factory

Issue: None

## Test plan:
- yarn test

### DIFF
--- a/flow-typed/ka-types.js
+++ b/flow-typed/ka-types.js
@@ -1,0 +1,8 @@
+// @flow
+
+// Marks all properties of an object optional.
+//
+// We use this instead of $Shape<T> because spreading $Shape<$Exact<T>> ends up
+// spreading $Exact<T> instead, losing the optionality of keys.
+// https://github.com/facebook/flow/issues/6906#issuecomment-453439922
+declare type Partial<T> = $ReadOnly<$Rest<T, {...}>>;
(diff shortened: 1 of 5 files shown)


flow-typed/ka-types.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+
+// Marks all properties of an object optional.
+//
+// We use this instead of $Shape<T> because spreading $Shape<$Exact<T>> ends up
+// spreading $Exact<T> instead, losing the optionality of keys.
+// https://github.com/facebook/flow/issues/6906#issuecomment-453439922
+declare type Partial<T> = $ReadOnly<$Rest<T, {...}>>;
```
kevinbarabash, line 8: I'm not sure if it makes sense to introduce a type like this in a library.

somewhatabstract: Do we still need it or can we use Flow's own `$Partial`?

kevinbarabash: I didn't realize that Flow had added that.  I'll switch to use that instead.

jeresig (approved):

PR #431, opened 2022-10-13
Correct flow types for isolateModules

somewhatabstract:
## Summary:

Issue: XXX-XXXX

## Test plan:
`yarn flow`
`yarn test`

### DIFF
--- a/packages/wonder-stuff-i18n/src/plugins/i18n-plugin.js
+++ b/packages/wonder-stuff-i18n/src/plugins/i18n-plugin.js
@@ -33,11 +33,11 @@ import type {TranslatedLocaleStrings} from "../utils/i18n-utils.js";
 /**
  * Add a new asset to the Webpack compilation assets.
  *
- * @param {Object} assets the compilation assets object from Webpack
+ * @param {Assets} assets the compilation assets object from Webpack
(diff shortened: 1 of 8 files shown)


packages/wonder-stuff-i18n/src/plugins/i18n-plugin.js
```diff
@@ -139,12 +139,15 @@ type InternalOptions = {
 type LocaleHashMap = {[oldHash: string]: string};
 type HashMaps = {[locale: string]: LocaleHashMap};
 
+type Asset = {|
+    source: () => string,
+    size: () => number,
+|};
```
kevinbarabash, line 145: **praise:** thanks for making this a separate type.

kevinbarabash (approved):
LGTM.  Thanks for tweak the type of `isolateModules`. 🎉

packages/wonder-stuff-i18n/src/plugins/i18n-plugin.js
somewhatabstract: It's odd how changing the isolateModules typing caused these other errors to show up, but it was worth fixing them :)

PR #502, opened 2023-01-23
TS Prep 1: Remove file extensions from all import statements

kevinbarabash:
## Summary:
This PR is not meant to land.  I'm experimenting with converting wonder-stuff to TS so that I can try out different ways to convert TS types to Flow so that we can continue using wonder-stuff in Flow projects after it's been migrated

Issue: FEI-4926

## Test plan:
- yarn lint

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -74,10 +74,7 @@ module.exports = {
         "import/no-named-default": "error",
         "import/extensions": [
             "error",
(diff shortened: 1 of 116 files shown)


kevinbarabash:
I'm going to ignore the test failures for now, but when we do this for real some of the snapshot tests will need updating.

PR #517, opened 2023-02-06
Use the 'v8' coverage provider for more accurate coverage reports

kevinbarabash:
## Summary:
When working on generating per-team coverage reports in webapp during the hackathon, John and I noticed how the total number of lines before reported in the reports was off by a lot, like two factors of magnitude.  By default jest uses the babel istanbul plugin to instrument the code, but it can miss lines, especially in JSX.  Switching to the "v8" coverage provider made a huge difference.  v8 computes coverage by tracking what JavaScript is run without making changes to the code itself.  This makes it more accurate, especially when it comes to JSX.  While JSX isn't a concern for wonder-stuff so it'll be interesting to see how much of an effect switching coverage providers will make.

Issue: None

## Test plan:
- let CI run
- look at top-line metrics to see how many more lines are being instrumented now


kevinbarabash:
After reviewing the changes in coverage, I don't think it makes sense to use the `v8` provider in this repo since there's no code using JSX here.

PR #523, opened 2023-02-13
FEI-4957.2: Replace enums with POJOs

kevinbarabash:
## Summary:
The codemod we're using to convert Flow to TypeScript doesn't support Flow enums.  This PR converts them to POJOs.  We'll convert them to TS enums in the "Cleanup" epic.  I've add a link to the ticket tracking that work for this repo.

Issue: FEI-4957

## Test plan:
- yarn lint
- yarn flow
- yarn test


jeresig (approved):

MiguelCastillo:
very excited to see this starting!

somewhatabstract (approved):
Looks good to me. Love seeing the FEI-4172 stuff getting removed at least.

> The codemod we're using to convert Flow to TypeScript doesn't support Flow enums.

It's a shame we can't fix the codemod so we don't have to do these steps.

kevinbarabash:
> It's a shame we can't fix the codemod so we don't have to do these steps.

We probably could, but it would be a lot of work since we have to convert each of the Flow enum runtime calls and remove the import statement for files using that runtime.  We aren't using Flow enums in that many places so investing in adding such complex logic to the codemod didn't seem like it was worth it.

PR #524, opened 2023-02-13
FEI-4957.3: Add tsconfig.json and library type definitions

kevinbarabash:
## Summary:
Before we can run the script to convert Flow to TypeScript, we need to have a tsconfig.json in place along with type definitions for as many of the third-party libraries that we're using as possible.  The next PR will contain the migrated files.

Issue: FEI-4957

## Test plan:
- let CI run


package.json
```diff
@@ -59,6 +64,7 @@
     "rollup-plugin-filesize": "^9.1.2",
     "rollup-plugin-preserve-shebangs": "^0.2.0",
     "rollup-plugin-terser": "^7.0.2",
+    "typescript": "^4.9.5",
```
MiguelCastillo, line 67: wooohoo!

package.json
```diff
@@ -71,6 +77,7 @@
     "clean": "rm -rf packages/wonder-stuff-*/dist && rm -rf packages/wonder-stuff-*/node_modules",
     "coverage": "yarn run jest --coverage",
     "flow:ci": "flow check --max-workers 0",
+    "format": "prettier --write .",
```
somewhatabstract, line 80: **question:** Why was this needed? Does this mean eslint will no longer apply prettier formatting in its autofixes?

tsconfig.json
```diff
@@ -0,0 +1,105 @@
+{
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig to read more about this file */
+
+    /* Projects */
```
somewhatabstract, line 5: **question:** Long term, do we need all these commented out things? From the perspective of a future maintainer, it's not clear to me if these are temporarily omitted things that used to be there or just helpful documentation - especially since they have values associated to them. Should we delete the commented out bits now and rely on the doc link for more info?

somewhatabstract (approved):
Some questions inline that I think need addressing (either with a change or at least a response to clarify things). Otherwise, it's exciting progress! 🎉

package.json
kevinbarabash: I forgot that I could've used `eslint --fix` to the formatting. 🙃

tsconfig.json
kevinbarabash: Probably not.  I ended up removing these in #528.  I'll re-add the link to https://aka.ms/tsconfig in that PR.

PR #526, opened 2023-02-14
FEI-4957.4: Run codemod to convert files to TypeScript

kevinbarabash:
## Summary:
This changeset was generated using the follow commands:
- cd ../flow-to-typescript-codemod
- yarn typescriptify convert -p ../wonder-stuff --write --delete
- // manually change 'export {* as jest} from "jest"' to be 'export * as jest from "jest"' in wonder-stuff
- yarn typescriptify fix --autoSuppressErrors --jiraSlug FEI-5011 -p ../wonder-stuff --config ../wonder-stuff/tsconfig.json
- // manually rename prettierrc.ts to prettierrc.js
- cd ../wonder-stuff
- yarn format

The next PR will fix all the TypeScript errors

Issue: FEI-4957

## Test plan:
- yarn tsc --noEmit, see a bunch of errors

```
Found 24 errors in 7 files.

Errors  Files
     2  packages/wonder-stuff-i18n/src/utils/emoji-for-locale.ts:46
    10  packages/wonder-stuff-sentry/src/normalize-sentry-data.ts:34
     2  packages/wonder-stuff-server-google/src/start-server.ts:26
     1  packages/wonder-stuff-server/src/middleware/default-error-logging.ts:18
     1  packages/wonder-stuff-server/src/middleware/default-request-logging.ts:18
     4  packages/wonder-stuff-server/src/start-server.ts:37
     4  packages/wonder-stuff-server/src/types.ts:1
```


jeresig (approved):
Amazing - that's really all the errors that remain? So few!

jeresig:
@kevinbarabash If you don't have a task for it yet, you're going to want to add a TS linter step that runs on a repo (probably have it check to see if a flowconfig exists, then run flow, and if a tsconfig exists, run TypeScript)

kevinbarabash:
> @kevinbarabash If you don't have a task for it yet, you're going to want to add a TS linter step that runs on a repo (probably have it check to see if a flowconfig exists, then run flow, and if a tsconfig exists, run TypeScript)

I don't have anything in the migration guide about setting up GitHub actions.  I'll add a section to call that out after I finish converting this repo.  It's all going to be under the same task.

packages/wonder-stuff-core/src/__tests__/clone-metadata.test.ts
```diff
@@ -34,15 +33,19 @@ describe("#cloneMetadata", () => {
             boolean: "true",
             undefined: undefined,
             null: null,
-        };
+        } as const;
 
         // Act
         const result = cloneMetadata(metadata);
+        // @ts-expect-error [FEI-5011] - TS2540 - Cannot assign to 'string' because it is a read-only property.
```
somewhatabstract, line 40: **note:** It's a shame we have all these error suppressions instead of just a judicial use of `any` on the variable from the beginning. Since these are test files, that feels like the right thing. Of course, that's manual work and would really take time, so I'm not requesting a change here. Just lamenting the tech debt/mess that is left behind.

packages/wonder-stuff-core/src/__tests__/clone.test.ts
```diff
@@ -37,7 +36,7 @@ describe("#clone", () => {
         // Arrange
         const originalValue = {
             array: [1, 2, 3],
-        };
+        } as const;
```
somewhatabstract, line 39: **question:** What is `as const` for? Why is it needed if the variable is  already `const`? Is it because TS demands that both sides of the `=` are const'd?

somewhatabstract (approved):
Good to see the automation do so much work for us. It's a shame that so many error suppressions get added when a judicial use of `any` or `mixed` would fix it (specifically regarding tests).

I am a little concerned by the suppressions in some of the more strongly-typed functions like `clone.js`, but we can revisit these things post-conversion, and they seem to be specific to the internals rather than leaking into the external representation, so no big deal.

packages/wonder-stuff-core/src/__tests__/clone.test.ts
kevinbarabash: I'm not sure why the tool adds `as const` in so many places.  We could change that behavior.  `as const` in this context changes the type that is inferred for the object literal.  With `as const`, it's inferred as precisely `{array: [1, 2, 3]}` whereas without it, it's inferred as `{array: number[]}`.  [TS playground example](https://www.typescriptlang.org/play?#code/MYewdgzgLgBCBOBLA5osBDANgNSwVwFMYBeGAbwCgZqZ1550BPALhgG0BGAGhgCYeAzAF0uFAL4UKoSLAQo0WXJkIBBCAGFw0EuSo06DFu259BI8bQgxp0ANxA)

kevinbarabash:
> I am a little concerned by the suppressions in some of the more strongly-typed functions like `clone.js`, but we can revisit these things post-conversion, and they seem to be specific to the internals rather than leaking into the external representation, so no big deal.

I think we might be able to solve this with conditional types to help feed more accurate type info through the body of the function so that we don't need the `// @ts-expect-error` comments.

packages/wonder-stuff-core/src/__tests__/clone-metadata.test.ts
kevinbarabash: I expect there will be quite a bit of post-conversion work to improve our types.

PR #528, opened 2023-02-14
FEI-4957.6: Use project references and multiple tsconfig.json files

kevinbarabash:
## Summary:
I ran into a number of issues with dts-bundle-generator.  It exports private that weren't exported and it can't handle having multiple types with the same name, e.g. different 'Options' types are defined in multiple places in wonder-stuff-server.

This PR address the issue by using project references which allow us to have multiple tsconfig.json files (one for each package) in a single repo.  Each tsconfig.json defines where the root (source) and output directories are relative to the tsconfig.json file.  The root tsconfig.json doesn't cover any source files and instead only references the other tsconfig.json in the package folders.

https://www.typescriptlang.org/docs/handbook/project-references.html#composite

Issue: FEI-4957

TODO:
- [x] update the existing build scripts to not generate a .d.ts file or .flow.js
- [x] update CI to run build:types

## Test plan:
- remove the `@ts-expect-error` comment from isolate-modules.ts, see that there's an error reported in VSCode
- `yarn clean`
- `yarn build:types`, see the same error reported
- `yarn typecheck`, see the same error
- re-add the the `@ts-expect-error` comment, see the error is gone in VSCode
- re-run the `yarn` commands from above, see that the error is no longer reported by them


jeresig (approved):
Ah ha! So it seems like we'll want to embrace the multiple config for all of our monorepos then.

Looks like there may be some flowtest files to update or remove!

kevinbarabash:
@jeresig request re-review due to the non-trivial changes I made to the tsconfig files.

.github/workflows/node-ci.yml
```diff
@@ -73,7 +73,11 @@ jobs:
       run: yarn build
 
     - name: Build Types
-      run: yarn build:types
+      # There isn't an easy way for us to type check our tests without
+      # also including them in dist folder when we build the .d.ts files
+      # so we have to manually remove them.
+      # TODO(kevin): Figure out how to have the `build:types` script do this.
+      run: yarn build:types && rm -rf packages/*/dist/**/__tests__
```
jeresig, line 80: Probably better to split this up onto multiple lines.

packages/wonder-stuff-server/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "exclude": ["dist"],
+  "exclude": ["dist", "**/*.flowtest.ts"],
```
jeresig, line 2: I'm fairly certain that typescript flowtest files make no sense.

jeresig (approved):
Ahh gotcha...

packages/wonder-stuff-server/tsconfig.json
kevinbarabash: I want to convert these to `.typetest.ts` files instead of deleting them. I'll do that in a separate PR.

.github/workflows/node-ci.yml
kevinbarabash: The removal of .d.ts files for tests gets rolled into the `build:types` script in a later PR.

PR #529, opened 2023-02-15
FEI-4957.7: Generate .js.flow files from .d.ts files

kevinbarabash:
## Summary:
We want to generate .js.flow files from our .d.ts files so that wonder-stuff packages can be used in codebases that are still using Flow.  This PR adds a build script to run flowgen on each .d.ts file in the dist folder of each package.

It also adds a script for removing .d.ts files that were generated from .test.ts files.  Both scripts are written in TypeScript and use swc-node/register to run.  I've added package.json scripts to make this easier.  I'll research this further to see if there's a nice shebang approach we can use in the future, but this should suffice for now.

Issue: FEI-4957

## Test plan:
- yarn clean
- yarn build:types
- yarn build:flowtypes
- See that .d.ts and .js.flow files have been added to each packages' dist folder
- See that there are no __tests__ folders in the dist folder
- Copy the dist folders for wonder-stuff-core and wonder-stuff-testing into the node_modules folder of wonder-blocks
- Restart flow in wonder-blocks, see that there are two Flow errors in wonder-blocks-icon (this is expected since the Flow types generated from the .d.ts files are different from the Flow types in the original code)


jeresig (approved):
Very nice! I bet we'll want to use that gen-flow-types.ts in wonder-blocks and perseus, too!

PR #530, opened 2023-02-15
FEI-4957.8: Remove Flow related plugins, settings, and lib defs

kevinbarabash:
## Summary:
This removes all Flow relate things.

NOTE: `eslint` is currently not running on .ts files.  I'll fix that in the next PR.

Issue: FEI-4957

## Test plan:
Let CI run the checks


packages/eslint-config-khan/index.js
```diff
@@ -72,30 +69,6 @@ module.exports = {
         "template-curly-spacing": OFF, // TODO(kevinb): enable this
         "valid-jsdoc": OFF, // TODO(kevinb): Enable this since we are using jdocs in some places.
 
-        /**
```
jeresig, line 75: Are there TS eslint rules we want to be enabling?

jeresig (approved):
Thanks for cleaning all of this up!

packages/eslint-config-khan/index.js
kevinbarabash: Yes, the migration guide has a mapping from Flow rules to TypeScript rules.  I'll add those to the config in the next PR when I enable eslint to check .ts files.

jeresig: Great, ok!

PR #531, opened 2023-02-15
FEI-4957.9: Update eslint config to use typescript parser and plugins

kevinbarabash:
## Summary:
It took a while to figure out the magic incantations to get imports to be recognized.  The config right now is split over packages/eslint-khan-config/index.js and .eslintrc.js.  In a separate PR, I'm going to pull stuff that looks like it's not repo-specific out of .eslintrc.js and into eslint-khan-config/index.js.  We can use the latter to help with migrating other codebases.

This PR also needed to make some changes to how we were mocking npm modules in our tests.  TBH, I'm not quite sure what changed in this PR that precipitated the need to make this change.  Previous PRs in the stack were running the tests and the test were passing fine then.  It may have been the change of some imports from default imports to namespace imports.

Issue: FEI-4957

## Test plan:
- yarn lint


packages/eslint-config-khan/index.js
```diff
@@ -4,9 +4,14 @@ const WARN = "warn";
 const ERROR = "error";
 
 module.exports = {
-    parser: "babel-eslint",
-    plugins: ["jsx-a11y", "prettier", "react"],
-    extends: ["eslint:recommended", "prettier"],
+    parser: "@typescript-eslint/parser",
+    plugins: ["@typescript-eslint", "jsx-a11y", "prettier", "react"],
```
jeresig, line 8: Will these changes impact other non-TS packages that are importing eslint-config?

tsconfig-common.json
```diff
@@ -8,7 +8,7 @@
         /* Modules */
         // Required for dynamic imports even though we aren't using
         // tsc to output any modules.
-        "module": "ESNext",
+        "module": "CommonJS",
```
jeresig, line 11: Is this comment still true?

jeresig (approved):
Awesome - thanks for figuring this all out!

packages/eslint-config-khan/index.js
kevinbarabash: It will, but only if they update to the new version.  It looks like wonder-block, perseus, and mobile/webview are using it.

tsconfig-common.json
kevinbarabash: I can revert this change.  I was experimenting with things.  I don't believe that the settings in this tsconfig affect the babel settings.

packages/eslint-config-khan/index.js
kevinbarabash: I'll be cleaning this up in the next PR since lots of things that should be in this file are actually in .eslintrc.js right now.

PR #534, opened 2023-02-16
FEI-4957.10: Re-add entries, keys, and values wrappers

kevinbarabash:
## Summary:
I did some more thinking about this and updating all of our uses of these wrappers in webapp is going to be a pain. It'll be easier to drop the wrappers after the conversion. I was able to add types for the overloads that TypeScript provides. It'll be interesting to see how well these translate to Flow.  I've also update the eslint config to ignore variables that are just underscores which is what we had originally.

Issue: FEI-4957

## Test plan:
- yarn lint
- yarn build:types
- yarn build:flowtypes
- copy/paste the types for entries, keys, and values into the Flow playground and do some spot check testing with them (it's okay if their behavior isn't exactly the same as the old Flow types for these wrappers)

### DIFF
--- a/.changeset/long-pears-laugh.md
+++ b/.changeset/long-pears-laugh.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-stuff-core": minor
+---
+
+Re-add entries, keys, and values wrappers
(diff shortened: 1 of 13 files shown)


jeresig (commented):
@kevinbarabash I think you forgot to add the files!

kevinbarabash:
I'll add `*.typestest.ts` to the list of files excluded from coverage.

PR #535, opened 2023-02-16
FEI-4957.11: Reorganize eslint configs

kevinbarabash:
## Summary:
When I initially updating the eslint config to support TypeScript, I added settings to .eslintrc.js that I should've added to eslint-config-khan/index.js.  Without these settings, using @khanacademy/eslint-config in other codebases would fail.

TODO:
- [x] update eslint-config-khan's package.json to have the appropriate peer deps

Issue: FEI-4957

## Test plan:
- add some lint
- see that it's reported in VSCode
- yarn lint, see that it's reported
- fix the lint
- yarn lint, see there's no lint

### DIFF
--- a/packages/eslint-config-khan/index.js
+++ b/packages/eslint-config-khan/index.js
@@ -27,20 +27,6 @@ module.exports = {
         "import/parsers": {
             "@typescript-eslint/parser": [".ts", ".tsx"],
         },
-        "import/resolver": {
-            typescript: {
-                project: [
-                    "packages/*/tsconfig.json",
-                    "packages/tsconfig-shared.json",
-                ],
-            },
-            node: {
-                project: [
-                    "packages/*/tsconfig.json",
-                    "packages/tsconfig-shared.json",
-                ],
-            },
-        },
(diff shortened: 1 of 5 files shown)


packages/eslint-config-khan/index.js
```diff
@@ -24,6 +24,24 @@ module.exports = {
         react: {
             version: "16.4",
         },
+        "import/parsers": {
+            "@typescript-eslint/parser": [".ts", ".tsx"],
+        },
+        "import/resolver": {
+            typescript: {
+                project: [
+                    "packages/*/tsconfig.json",
```
jeresig, line 33: Does the `packages` pattern work on other repos? And in webapp?

jeresig (approved):

packages/eslint-config-khan/index.js
kevinbarabash: Thanks for catching that.  That is definitely repo specific.  I'll move that back to .eslintrc.js and update the README.md to include a note about this.

PR #546, opened 2023-02-21
[🔥AUDIT🔥] Update wonder-stuff-core to export 'entries', 'keys', and 'values' wrapper functions

kevinbarabash:
🖍 _This is an audit!_ 🖍

## Summary:
I had removed them in one of the PRs while migrating wonder-stuff to TypeScript and then I thought better of it and re-added them, but forgot to update index.ts so they were never exported by the package.  This audit fixes that.

Issue: None

## Test plan:
- yarn build
- check the build product to see that they're included and exported

### DIFF
--- a/.changeset/sweet-lemons-argue.md
+++ b/.changeset/sweet-lemons-argue.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-stuff-core": minor
+---
+
+Export 'entries', 'keys', and 'values' wrappers
--- a/packages/wonder-stuff-core/src/index.ts
+++ b/packages/wonder-stuff-core/src/index.ts
@@ -1,10 +1,13 @@
 export {clone} from "./clone";
+export {entries} from "./entries";
 export {Errors} from "./errors";
 export {errorsFromError, Order} from "./errors-from-error";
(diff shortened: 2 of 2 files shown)


jeresig (approved):

.github/workflows/node-ci.yml
```diff
@@ -84,11 +84,11 @@ jobs:
...
         full-trigger: ${{ steps.eslint-reset.outputs.filtered }}
         full: yarn lint:ci
-        limited-trigger: ${{ steps.js-files.outputs.filtered }}
+        limited-trigger: ${{ steps.js-ts-files.outputs.filtered }}
         limited: yarn lint:ci {}
 
     - name: Typecheck
-      if: steps.js-files.outputs.filtered != '[]'
+      if: steps.js-ts-files.outputs.filtered != '[]'
       run: yarn typecheck
```
kevinbarabash, line 92: I noticed this step wasn't running when reviewing the checks.  Upgrading the `js-files` step to `js-ts-files` should fix the issue.

kevinbarabash: I've verified the fix by looking at the logs from the most recent workflow run.

PR #567, opened 2023-03-08
Migrate more render-gateway pieces to Wonder Stuff

somewhatabstract:
## Summary:
The aim here was to migrate into Wonder Stuff the work that the `ka-shared` fake package in the render-gateway repo is doing. However, it doesn't really map 1-1. Some of the code will ultimately go into the render-gateway Wonder Stuff package or the webapp service rather than Wonder Stuff Server package. Instead, this migrates pieces from various parts of render-gateway that ultimately would be useful within a generic "server" package that we could use for any server, such as request authentication or common APIs.

So, this migrates the request authentication middleware, common service routes middleware, and buffer decryption via KMS over to Wonder Stuff Server.

It also consolidates Wonder Stuff Server Google and Wonder Stuff Server into a single package. While having things separate long term would be great for enabling us to shift to alternative hosts, it's a bigger lift right now to get the architecture right. We can revisit that another time, post-migration, as needed.

Next, I'll be adding a new package, Wonder Stuff Render Server, and then migrating the remaining render-gateway repo pieces into that package or the existing Wonder Stuff Server package as needed. After that, it'll be time to reimplement the Webapp render-gateway service using these new packages.

This also adds the beginnings of secrets support to Wonder Stuff Core. I considered using a new Wonder Stuff Secrets package for this, but I wasn't convinced it needed it's own thing. Long term, I hope to bring over some additional secrets support around the secrets config files we have and the tooling that we use to turn those into the secrets from Keeper or other sources. That's outside the scope of this work, but perhaps adds some context for reviewers as to why a separate Wonder Stuff Secrets package might be valuable in the future.

Finally, this removes the `@types/winston` and `@types/winston-express`. I noticed when running yarn that these are recommended to be removed as they are considered stubs and the real packages contain the types.

Issue: FEI-4005

## Test plan:
`yarn test`
`yarn typecheck`


jeresig (approved):
Nice - didn't investigate it in detail, I assume this is mostly moving stuff around and doing some TS conversion, but seems ok to me!

PR #606, opened 2023-03-30
[fei4960.1.rendergateway] Add migrated render server package

somewhatabstract:
This adds two new packages to Wonder Stuff:
  - `@khan-academy/wonder-stuff-render-server`
  - `@khan-academy/wonder-stuff-render-environment-jsdom`

The first is a complete migration of the main functionality of Khan/render-gateway over to TypeScript. Still pending are some of the examples that can help verify it is working. These will be added in an upcoming PR.

The second is a placeholder for the JSDOM render environment. This will be migrated from Khan/render-gateway's `environments/jsdom` folder in the next PR.

This update also adds a new type export, Mutable<T>, to Wonder Stuff Core that can be used to make a readonly into a mutable version of itself.

Finally, there are some eslint configuration tweaks to make it easier to add `jest-extended` to test files, as well, as some removed type suppressions after doing so.

Issue: FEI-4960

## Methodology:
I followed our migration steps per the Confluence documentation except for the part to suppress errors. I took a moment to look into the errors and realised `jest-extended` was needed. So I added imports for that and fixed the eslint issues. I then noticed the rest were mostly casting issues or incorrect types since there weren't 1:1 mappings - so I fixed them. No suppressions needed. 

## Test plan:
`yarn test`
`yarn build`
`yarn typecheck`

PR #607, opened 2023-03-30
[fei4960.2.jsdom] Migrate JSDOM render environment

somewhatabstract:
## Summary:
This migrates the JSDOM render environment from Khan/render-gateway. Now it is its own package as it should've been.

This also tweaks the rollup config to cope with dynamic imports which was needed for the new package to build.

The only migration piece left is to bring in the example server implementations for testing various paths through the code.

Issue: FEI-4960

## Test plan:
`yarn test`
`yarn build`
`yarn typecheck`


build-settings/rollup.config.js
```diff
@@ -103,8 +113,9 @@ const createConfig = (
 
     const extensions = [".js", ".jsx", ".ts", ".tsx"];
 
+    const isSingleFile = inputFile != null;
     const config = {
-        output: createOutputConfig(name, format, file),
+        output: createOutputConfig(name, format, file, isSingleFile),
         input: makePackageBasedPath(name, inputFile || "./src/index.ts"),
```
kevinbarabash, line 119: **suggestion (non-blocking):** we default to using `./src/index.ts` in the case that there is no `inputFile`.  It seems like there's still a single file that we're compiling in that case.  I think some of these variable names could be reworked (or maybe a new property on the config object that gets passed in) to help make it easier to understand what's going on.

packages/wonder-stuff-render-environment-jsdom/tsconfig.json
```diff
@@ -7,5 +7,9 @@
     "outDir": "dist",
     "rootDir": "src",
   },
-  "references": []
+  "references": [
+    {"path": "../wonder-stuff-core"},
+    {"path": "../wonder-stuff-server"},
+    {"path": "../wonder-stuff-render-server"},
+  ]
```
kevinbarabash, line 14: **suggestion:** you may need to update tsconfig-build.json to include a reference to `wonder-stuff-render-environment-jsdom` so that `.d.ts` files are built for this package.

kevinbarabash (approved):
LGTM.  Thanks for figuring out the issue with the build.  I've left a few comments, but nothing blocking.

build-settings/rollup.config.js
somewhatabstract: It's not about the input file, it's about the outputs.

packages/wonder-stuff-render-environment-jsdom/tsconfig.json
somewhatabstract: Oh good catch. I've done that here (and for the previous PR - so I won't fix it there).

somewhatabstract:
The parent pull-request ([#606](https://github.com/Khan/wonder-stuff/pull/606)) has been merged into `main`, but this branch (`fei4960.2.jsdom`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #616, opened 2023-04-03
[🔥AUDIT🔥] [rg.audit.1] Add custom flow file for render-environment-jsdom to work around flowgen issue

somewhatabstract:
🖍 _This is an audit!_ 🖍

## Summary:
- Add the faster flowgen script from Wonder Blocks, and include the "custom overrides" support.
- Fix some flow type issues for core and render-environment-jsdom.

Issue: FEI-4960

## Test plan:
`yarn clean && yarn build && yarn build:types && yarn build:flowtypes`

Also check that the types will work for the usages I have to hand.


build-scripts/gen-flow-types.ts
somewhatabstract, line 1: Copied from Wonder Blocks, but edited to:

- Use the correct path for `packages/wonder-stuff-*`
- Remove React-specific tweaks

PR #620, opened 2023-04-03
[🔥AUDIT🔥] [rg.audit.3] Fix keys typing

somewhatabstract:
🖍 _This is an audit!_ 🖍

## Summary:
This updates the typing of the `keys` function to match better with our expected usage.

Issue: FEI-4960

## Test plan:
`yarn typecheck`

Also built the types and tried them in webapp to verify they now work as we would want.


kevinbarabash (commented):
Thanks for fixing this. 💚

PR #628, opened 2023-04-06
Creating wonder-stuff-ci package!

jlauzy:
## Summary:
- Created a new package `wonder-stuff-ci`. This package will contain typescript functions that are used throughout our scripts and automations ( OLC, Bitrise, Github Actions, etc.). I found a need for this when needing common functions for interacting with mobile releases.
- Ported over three functions that I need and will continue adding functions into this package as they arise.

Issue: FEI-5058

## Test plan:
- Wrote Jest test for the functions.


packages/wonder-stuff-ci/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# @khanacademy/wonder-stuff-ci
+
+## 0.0.1
+
+- Initial commit of porting over OLC functions for mobile release management
```
jlauzy, line 5: How should I format the change log in wonderstuff? Is there a getting started guide on creating a package in wonder-stuff? All I did was copy what another package was doing.

packages/wonder-stuff-ci/package.json
```diff
@@ -0,0 +1,20 @@
...
+        "node": ">=16"
+    },
+    "name": "@khanacademy/wonder-stuff-ci",
+    "version": "0.0.1",
+    "description": "Functions for creating automation and scripts.",
+    "scripts": {
+        "test": "bash -c 'yarn --silent --cwd \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
+    },
+    "devDependencies": {
+        "@types/node": "^18.15.11",
```
jlauzy, line 15: I needed to add this to resolve some types in the functions i need.

packages/wonder-stuff-ci/package.json
```diff
@@ -0,0 +1,20 @@
...
+        "ws-dev-build-settings": "^1.1.0"
```
jlauzy, line 16: What is the ws-dev-build-settings? Should I be adding the @types/node into this?

packages/wonder-stuff-ci/package.json
```diff
@@ -0,0 +1,20 @@
...
+        "access": "public"
+    },
+    "engines": {
+        "node": ">=16"
+    },
+    "name": "@khanacademy/wonder-stuff-ci",
+    "version": "0.0.1",
+    "description": "Functions for creating automation and scripts.",
+    "scripts": {
+        "test": "bash -c 'yarn --silent --cwd \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
```
jlauzy, line 12: I was trying to get the test to run with `yarn jest`, but that didn't work because of typing. But using the yarn command it did work!

tsconfig-build.json
```diff
@@ -12,5 +12,6 @@
     {"path": "./packages/wonder-stuff-sentry"},
     {"path": "./packages/wonder-stuff-server"},
     {"path": "./packages/wonder-stuff-testing"},
+    {"path": "./packages/wonder-stuff-ci"},
```
jlauzy, line 15: I was curious how the tsconfig-build.json is used in wonderstuff?

What else is needed for me to get this new package published to npm?

jlauzy:
@somewhatabstract @kevinbarabash , uh did i do this right lol

packages/wonder-stuff-ci/CHANGELOG.md
jeresig: Ahh no - so we're using Changesets! The CHANGELOG is generated automatically, so you shouldn't make it. You can see how to do this here: https://khanacademy.atlassian.net/wiki/spaces/WB/pages/2217640195/Versioning+and+releasing+Wonder+Blocks+packages (it's Wonder Blocks but we follow the same process for Wonder Stuff)

packages/wonder-stuff-ci/package.json
jeresig: This is internal to wonder-stuff itself, these are the shared tools that we use to help with test running, builds, etc. I'm not sure if it makes sense to have the node types, or not! I'd say "no" only because not everything in wonder-stuff is designed to run in Node.

packages/wonder-stuff-ci/package.json
jeresig: Ah yeah - I think this handles some of the trickiness with being in a monorepo, if I remember correctly.

packages/wonder-stuff-ci/src/release-utils.ts
```diff
@@ -0,0 +1,70 @@
+import execProm from "./exec-prom";
```
jeresig, line 1: I would rename this file to `mobile-release-utils.ts`. That being said, this logic is rather specific to our release processes!

tsconfig-build.json
jeresig: All of it is managed by Changesets - once you changes are approved and landed then a Changesets PR will be generated, once that's approved and landed it'll publish to NPM!

packages/wonder-stuff-ci/package.json
```diff
@@ -0,0 +1,20 @@
...
+    "publishConfig": {
+        "access": "public"
+    },
+    "engines": {
+        "node": ">=16"
+    },
+    "name": "@khanacademy/wonder-stuff-ci",
+    "version": "0.0.1",
+    "description": "Functions for creating automation and scripts.",
+    "scripts": {
```
jeresig, line 11: I think you're missing a critical `index.ts` file with all the things that you've created exported. Looking at wonder-stuff-core you'll probably want to add these lines here:

```
    "module": "dist/es/index.js",
    "main": "dist/index.js",
    "types": "dist/index.d.ts",
```

And then make a `src/index.ts` file that imports your files and re-exports them again.

packages/wonder-stuff-ci/src/release-utils.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+    return 0;
+};
+
+/**
+ * Matches [release/[unified|ios|android]]/[v]<num>.<num>.<num>[-extra]
+ * @param arg
+ * @returns match, if found, null otherwise
+ */
+export const isGetReleaseBranch = (arg: string) => {
```
jeresig, line 42: I would rename this to: `extractMobileReleaseBranch`.

packages/wonder-stuff-ci/src/release-utils.ts
```diff
@@ -0,0 +1,70 @@
...
+    return match && match.length >= 3 && match[3]
+        ? {prefix: match[1] || "release/unified/", version: match[3]}
+        : null;
+};
+
+/**
+ *
+ * @returns all release tags sorted creation time ascending
+ */
+const allReleaseTags = async (): Promise<Array<string>> => {
```
jeresig, line 59: I would re-name this to: `allMobileReleaseTags`

jeresig (changes requested):
Exciting! Suggestions and corrections inline.

packages/wonder-stuff-ci/src/exec-prom.ts
```diff
@@ -0,0 +1,41 @@
...
+            err
+                ? rejectOnError
+                    ? rej(err)
+                    : res({
+                          err,
+                          stdout: bufferToString(stdout),
+                          stderr: bufferToString(stderr),
+                      })
+                : res({
+                      err: null,
```
kevinbarabash, line 34: **suggestion (non-blocking):** You could use `null` here which would allow the type for `err` to be `Error | undefined`.

packages/wonder-stuff-ci/src/exec-prom.ts
```diff
@@ -0,0 +1,41 @@
...
+ * A simple promisified version of child_process.exec, so we can `await` it
+ */
+import {exec, ExecOptions} from "child_process";
+
+export const bufferToString = (input: Buffer | string): string => {
+    if (typeof input === "string") {
+        return input;
+    } else {
+        return input.toString("utf8");
+    }
```
kevinbarabash, line 11: **suggestion (non-blocking):** The following would be a bit more succinct:
```suggestion
    return (typeof input === "string") ? input : input.toString("utf8");
```

packages/wonder-stuff-ci/src/mobile-release-utils.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Compares two versions of the form <num>.<num>.<num>
+ * @param v1
+ * @param v2
+ * @returns int, 1 if v1 > v2, -1 if v1 < v2, 0 if v1 == v2
+ */
+export const compareVersions = (v1: string, v2: string) => {
+    const v1v = v1.includes("-") ? v1.split("-")[1] : v1;
+    const v2v = v2.includes("-") ? v2.split("-")[1] : v2;
```
kevinbarabash, line 24: **question:** why are we checking for `"-"` here?  The docstring says the version should be of the form `<num>.<num>.<num>`.

packages/wonder-stuff-ci/src/mobile-release-utils.ts
```diff
@@ -0,0 +1,70 @@
...
+ * @returns match, if found, null otherwise
+ */
+export const extractMobileReleaseBranch = (arg: string) => {
+    if (!arg) {
+        return null;
+    }
+
+    const match = arg.match(
+        /^(release\/(ios|android|unified)\/)?v?(\d+\.\d+\.\d+(-\w*)*)$/i,
+    );
```
kevinbarabash, line 49: **praise:** I like that we're using a regex for this.  Maybe we can do the same for `compareVersions` so that there's more consistency with how we're parsing version strings.  This would allow the code in that function to be less defensive (we could assume there's always three numbers in a version string).

packages/wonder-stuff-ci/src/__tests__/mobile-release-utils.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        expect(result).toBe(0);
+    });
+});
+
+describe("#isGetReleaseBranch", () => {
+    it.each([
+        "release/unified/7.8.0",
+        "release/android/7.8.0",
+        "release/ios/7.8.0",
+    ])("get the version from the release branch", (testCase: any) => {
```
kevinbarabash, line 55: **question:** does TS complain if we type the `testCase` as `string` instead of `any`?

kevinbarabash (approved):
LGTM.  A few nits, but nothing blocking.  Thanks for creating this new shared package.

packages/wonder-stuff-ci/src/exec-prom.ts
jlauzy: Agree!

packages/wonder-stuff-ci/src/mobile-release-utils.ts
jlauzy: Ah, I need to update the docstring! tags are in the form `unified-7.9.0`, but this function also takes in raw version string `7.9.0`

packages/wonder-stuff-ci/src/mobile-release-utils.ts
jlauzy: Ah I see what you are saying 
```
    const v1p = v1v.replace(/^v/g, "").split(".");
    const v2p = v2v.replace(/^v/g, "").split(".");
    for (let i = 0; i < v1p.length || i < v2p.length; i++) {
        const p1 = +v1p[i] || 0;
        const p2 = +v2p[i] || 0;
        if (+p1 !== +p2) {
            return p1 > p2 ? 1 : -1;
        }
    }
 ```
 This does not check if there are 3 numbers in the version string only. It just checks each digit that is split by "."

File: packages/wonder-stuff-ci/src/__tests__/mobile-release-utils.test.ts (hunk 14 above)
Comment (jlauzy) [reply packages/wonder-stuff-ci/src/__tests__/mobile-release-utils.test.ts:57]: hmm interestingly my IDE does not complain. I ran `tsc` no complaints. I even ran `yarn test` and no complaints.

**Review by jeresig — APPROVED** [review]
Thanks for the changes!

File: packages/wonder-stuff-ci/README.md [hunk 15]
```diff
@@ -0,0 +1,3 @@
+# wonder-stuff-ci
+
+This Wonder Stuff package contains functions for creating automation and scripts.
```
somewhatabstract, line 3: **nit:**
```suggestion
This Wonder Stuff package contains functions for automation and scripts.
```

packages/wonder-stuff-ci/package.json
somewhatabstract: Typically, you have to run `yarn jest` and `yarn test` from the root directory. This confusing command at the package level basically does that for you, so you can do `yarn test` in one of the package folders, and it will do the right thing.

packages/wonder-stuff-ci/src/mobile-release-utils.ts
```diff
@@ -0,0 +1,54 @@
...
+            tag.match(/^(ios|android|unified)-(\d+\.\d+\.\d+(-\w*)*)$/i),
+        )
+        .sort(compareVersions);
+};
+
+/**
+ * Compares two versions (of the form <num>.<num>.<num> ) or tags (of the form <tag>-<num>.<num>.<num>)
+ * @param v1
+ * @param v2
+ * @returns int, 1 if v1 > v2, -1 if v1 < v2, 0 if v1 == v2
```
somewhatabstract, line 21: **suggestion:** We use the imperative for function names, and try to keep them under 80 chars in width. Longer descriptive info can go underneath. In addition, the params and returns should have type info and descriptions.

```suggestion
 * Compare two versions or tags
 * @param {string} v1 A version of the form `<num>.<num>.<num>` or tag of the form `<tag>-<num>.<num>.<num>`.
 * @param {string} v2 A version of the form `<num>.<num>.<num>` or tag of the form `<tag>-<num>.<num>.<num>`.
 * @returns {number} 1 if v1 > v2, -1 if v1 < v2, 0 if v1 == v2
```

packages/wonder-stuff-ci/src/mobile-release-utils.ts
```diff
@@ -0,0 +1,54 @@
+import execProm from "./exec-prom";
+import { allMobileReleaseTags } from "./mobile-release-git-utils";
+
+/**
+ *
+ * @returns all release tags sorted creation time ascending
```
somewhatabstract, line 6: **suggestion:** Don't forget a first docline as well as types on the return.
```suggestion
 * Get mobile release tags.
 *
 * Tags are filtered to only include those matching our tag version format
 * (`<tag>-<num>.<num>.<num>`), then they are sorted by the version information from
 * earliest version to most recent.
 *
 * @returns {Promise<Array<string>>} all release tags sorted creation time ascending
```

packages/wonder-stuff-ci/src/mobile-release-utils.ts
```diff
@@ -0,0 +1,54 @@
...
+            return p1 > p2 ? 1 : -1;
+        }
+    }
+    return 0;
+};
+
+/**
+ * Matches [release/[unified|ios|android]]/[v]<num>.<num>.<num>[-extra]
+ * @param arg
+ * @returns match, if found, null otherwise
```
somewhatabstract, line 41: **suggestion:** Same again. Imperative, as well as updating the params and returns.
```suggestion
 * Extract release branch from tag.
 *
 * @param {string} The tag of the form `[release/[unified|ios|android]]/[v]<num>.<num>.<num>[-extra]`.
 * @returns {string} The matched branch, if found; otherwise, `null`.
```

packages/wonder-stuff-ci/src/mobile-release-git-utils.ts
```diff
@@ -0,0 +1,14 @@
+import execProm from "./exec-prom";
+
+/**
+ *
+ * @returns all release tags sorted creation time ascending
```
somewhatabstract, line 5: **suggestion:** We use the imperative to describe a function, as well as type the returned values (and params, as necessary).
```suggestion
 * Get all mobile release tags.
 * 
 * @returns {Promise<Array<string>>} A promise of all release tags sorted by creation time ascending.
```

Also, this seems like it could very easily be turned into a more generic, `getAllMatchingTags` by moving the `ios|android|unified` filter to an argument. No change requested - just noting that if this starts needing to get reused in non-mobile contexts, we may see the need to change that.

packages/wonder-stuff-ci/src/__tests__/exec-prom.test.ts
```diff
@@ -0,0 +1,18 @@
+import {buffer} from "stream/consumers";
+import {execProm, bufferToString} from "../exec-prom";
+
+describe("#bufferToString", () => {
+    it.each(["testing", Buffer.from("testing")])(
+        "the buffer to string function returns the correct value",
+        (testCase: any) => {
```
somewhatabstract, line 7: **nit:**
```suggestion
        (testCase: string | Buffer) => {
```

packages/wonder-stuff-ci/src/__tests__/mobile-release-git-util.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+jest.mock("../exec-prom");
+
+describe("#allMobileReleaseTags", () => {
+    it("return an array of release tags split by new line", async () => {
+        // Arrange
+        jest.spyOn(require("../exec-prom"), "execProm").mockReturnValue({
+            stdout: "android-7.10.0\nunified-7.8.0\nunified-7.9.0",
+        });
+        // Act
```
somewhatabstract, line 12: **nit:**
```suggestion
        });

        // Act
```

packages/wonder-stuff-ci/src/__tests__/mobile-release-utils.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    compareVersions,
+    extractMobileReleaseBranch,
+    getMobileReleaseTags,
+} from "../mobile-release-utils";
+import {allMobileReleaseTags} from "../mobile-release-git-utils";
+
+describe("#compareVersions", () => {
+    it.each(["7.8.1", "7.9.0", "7.10.0", "8.0.0"])(
+        "return 1 if version 1 is greater than version 2",
+        (testCase: any) => {
```
somewhatabstract, line 11: **nit:**
```suggestion
        (testCase: string) => {
```

packages/wonder-stuff-ci/src/__tests__/mobile-release-utils.test.ts
```diff
@@ -0,0 +1,112 @@
...
+            const result = compareVersions(version1, version2);
+
+            // Assert
+            expect(result).toBe(1);
+        },
+    );
+
+    it.each(["7.7.1", "7.6.0", "6.10.0"])(
+        "return -1 if version 1 is less than version 2",
+        (testCase: any) => {
```
somewhatabstract, line 26: **nit:**
```suggestion
        (testCase: string) => {
```

packages/wonder-stuff-ci/src/__tests__/mobile-release-utils.test.ts
```diff
@@ -0,0 +1,112 @@
...
+        expect(result).toBe(0);
+    });
+});
+
+describe("#isGetReleaseBranch", () => {
+    it.each([
+        "release/unified/7.8.0",
+        "release/android/7.8.0",
+        "release/ios/7.8.0",
+    ])("get the version from the release branch", (testCase: any) => {
```
somewhatabstract, line 57: **nit:**
```suggestion
    ])("get the version from the release branch", (testCase: string) => {
```

packages/wonder-stuff-ci/src/__tests__/mobile-release-utils.test.ts
```diff
@@ -0,0 +1,112 @@
...
+        // Act
+        const result = extractMobileReleaseBranch(releaseBranch);
+
+        // Assert
+        expect(result?.version).toBe("7.8.0");
+    });
+
+    it.each(["release/testing", "android/7.8.0", "ios/7.8.0"])(
+        "return null if the branch is not a release branch",
+        (testCase: any) => {
```
somewhatabstract, line 70: **nit:**
```suggestion
        (testCase: string) => {
```

packages/wonder-stuff-ci/src/exec-prom.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * A simple promisified version of child_process.exec, so we can `await` it
+ */
+import {exec, ExecOptions} from "child_process";
+import util from "util";
+
+export const bufferToString = (input: Buffer | string): string => {
```
somewhatabstract, line 7: **suggestion:** Can you add a jsdoc comment block here?
```suggestion
/**
 * Coerce a buffer or string into a string.
 *
 * @param {Buffer | string} input The `Buffer` or `string`
 * that should be coerced to a string.
 * @returns {string} The `string` representation of the given parameter.
 */
export const bufferToString = (input: Buffer | string): string => {
```

packages/wonder-stuff-ci/src/exec-prom.ts
```diff
@@ -0,0 +1,13 @@
...
+    return typeof input === "string" ? input : input.toString("utf8");
+};
+
+export const execProm = util.promisify(exec);
```
somewhatabstract, line 11: **suggestion:** Generally, when promisifying async methods, I have used the convention of adding `Async` to the name. It is generally clearer to grasp and understand what the function is for in that scenario.

That said, I'm not sure it's worth the abstraction of its own file rather than just creating the method in the files that use it - except maybe to simplify testing?

When you change the name of the function, don't forget to change the name of the file to match.

```suggestion
export const execAsync = util.promisify(exec);
```

packages/wonder-stuff-ci/src/exec-prom.ts
somewhatabstract, line 7: **suggestion:** Generally, we keep each file to one export. It makes discoverability easier and we have no quota on files so there's little reason not to do it. As it stands, it's kinda confusing that the `exex-prom.ts` file exports `bufferToString`. Can you move `bufferToString` to `buffer-to-string.ts`?

packages/wonder-stuff-ci/src/mobile-release-utils.ts
somewhatabstract, line 1: **suggestion:** The general convention is to provide one export per file. It makes it easier to find things and organize things, and often eases maintenance. Could you move each of these functions to their own files with their own tests?

somewhatabstract (changes requested):
This is looking good. I have a few changes I'd like to see before this gets landed. See my inline comments for info.

jlauzy:
@somewhatabstract Thanks for the feedback jeff, updated the pr based on your comments. Let me know if I missed anything!

packages/wonder-stuff-ci/package.json
```diff
@@ -7,7 +7,7 @@
     },
     "name": "@khanacademy/wonder-stuff-ci",
     "version": "0.0.1",
-    "description": "Functions for creating automation and scripts.",
+    "description": "Functions for automation and scripts.",
```
somewhatabstract, line 10: **praise:** Thanks!

packages/wonder-stuff-ci/src/extract-mobile-release-info-from-branch-name.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Example, given the branch `release/unified/v7.8.0`:
+ * {
+ *    prefix: "release/unified/",
+ *    version: "7.8.0"
+ * }
+ *
+ * @param {string} The release branch of the form `[release/[unified|ios|android]]/[v]<num>.<num>.<num>[-extra]`.
+ * @returns {Object} The release version and prefix, if found; otherwise, `null`.
+ */
+export const extractMobileReleaseInfoFromBranchName = (arg: string | null) => {
```
somewhatabstract, line 13: **suggestion:** Currently, the return type is inferred from the code and as such, there isn't a concrete way to refer to this type. It would be helpful for calling code if there was one as it can be used by them to type their own arguments and variables.

Something like:

```
type MobileReleaseInfo = {
    prefix: "release/unified/" | "release/ios" | "release/android",
    version: string,
};
```

And then:
```suggestion
export const extractMobileReleaseInfoFromBranchName =
    (arg: string | null): MobileReleaseInfo => {
```

packages/wonder-stuff-ci/src/extract-mobile-release-info-from-branch-name.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Extract release version and prefix from mobile release branch.
+ *
+ * Example, given the branch `release/unified/v7.8.0`:
+ * {
+ *    prefix: "release/unified/",
+ *    version: "7.8.0"
+ * }
+ *
+ * @param {string} The release branch of the form `[release/[unified|ios|android]]/[v]<num>.<num>.<num>[-extra]`.
+ * @returns {Object} The release version and prefix, if found; otherwise, `null`.
```
somewhatabstract, line 11: **suggestion:**  If you create the type as suggested below and then replace `{Object}` with `{MobileReleaseInfo}`, which gives folks a little nicer info in their IDEs.

somewhatabstract (approved):
Thanks for all these changes. This is looking great. I have one more change to suggest, but not blocking. See the inline commentary for info.

packages/wonder-stuff-ci/src/types.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * The information about a release branch.
+ */
+type MobileReleaseBranchInfo = {
```
somewhatabstract, line 4: **suggestion:** Don't forget to export this type from `index.ts` so consumers of the API can use it.

`export type {MobileReleaseBranchInfo} from "./types";`

or

`export * from "./types";`

packages/wonder-stuff-ci/src/extract-mobile-release-info-from-branch-name.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Extract release version and prefix from mobile release branch.
+ *
+ * Example, given the branch `release/unified/v7.8.0`:
+ * {
+ *    prefix: "release/unified/",
+ *    version: "7.8.0"
+ * }
+ *
+ * @param {string} The release branch of the form `[release/[unified|ios|android]]/[v]<num>.<num>.<num>[-extra]`.
+ * @returns {MobileReleaseInfo} The release version and prefix, if found; otherwise, `null`.
```
somewhatabstract, line 11: **suggestion:**
```suggestion
 * @returns {MobileReleasBranchInfo} The release version and prefix, if found; otherwise, `null`.
```

PR #636, opened 2023-04-12
Added a logger for sending logs to GCP

jlauzy:
## Summary:
Created a logger that will be used throughout the release process in order to create an audit log trail in GCP.

Issue: FEI-5057

## Test plan:
How the console log looks like
image

How logs look in GCP (we might want to increase our log retention in GCP from 30 days to something higher?)
image


packages/wonder-stuff-ci/src/get-mobile-release-logger.ts
```diff
@@ -0,0 +1,79 @@
...
+ * Winston logger used for generating to GCP for the mobile release process (release-raccoon).
+ * @param {string} projectId The GCP project ID to send logs to.
+ * @param {string} logName The name of the log to send logs to in GCP.
+ * @param {string} logLevel The minimum severity level to log.
+ * @param {boolean} enableConsoleLogs If true, logs will be sent to the console as well as GCP.
+ * @param {[key: string]: string} labels K/V metadata that will be add to the logs
+ * @returns {winston.Logger} logger set up to send logs to GCP.
+ */
+export const getMobileReleaseLogger = (
+    args = {
```
jlauzy, line 15: Is there a way to create default values for a parameters and then the caller passes only one value i.e. `enableConsoleLogs` set to true but everything else I want default?

packages/wonder-stuff-ci/src/get-mobile-release-logger.ts
```diff
@@ -0,0 +1,79 @@
...
+        projectId: "mobile-365917",
+        logName: "release-raccoon",
+        logLevel: "info",
+        enableConsoleLogs: false,
+        labels: {},
+    },
+): winston.Logger => {
+    // Set up logging configuration to GCP
+    const loggingGCPWinston = new gcpLogger.LoggingWinston({
```
jeresig, line 24: Hmm - so we already have some Winston logging logic here: https://github.com/Khan/wonder-stuff/blob/main/packages/wonder-stuff-server/src/get-logging-transport.ts

I wonder if it makes sense to combine these efforts?

packages/wonder-stuff-ci/src/get-mobile-release-logger.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @returns {winston.Logger} logger set up to send logs to GCP.
+ */
+export const getMobileReleaseLogger = (
+    args = {
+        projectId: "mobile-365917",
+        logName: "release-raccoon",
+        logLevel: "info",
+        enableConsoleLogs: false,
+        labels: {},
+    },
```
jeresig, line 21: To get the default values I would do this:
```suggestion
    {
        projectId = "mobile-365917",
        logName = "release-raccoon",
        logLevel = "info",
        enableConsoleLogs = false,
        labels = {},
    },
```

And then you would reference them by just saying `logLevel` instead of `args.logLevel`.

packages/wonder-stuff-ci/src/get-mobile-release-logger.ts
jlauzy: This is great! I'll definitely combine the efforts. The only change needed is to add this new GCP logging transport and then I don't have to create the logger and all the other stuff in this file.

packages/wonder-stuff-ci/src/get-mobile-release-logger.ts
jlauzy: Oh nice! Strange thing is in my IDE it complains if I pass nothing and just want the default values 
image

jeresig: Ah right - so you would have to have a *second* fallback in that case - you could provide that or just pass in an object object `{}` as the first argument.

packages/wonder-stuff-ci/src/get-mobile-release-logger.ts
```diff
@@ -0,0 +1,79 @@
...
+import * as gcpLogger from "@google-cloud/logging-winston";
+const {combine, timestamp, json, errors, printf, colorize} = winston.format;
+
+/**
+ * Winston logger used for generating to GCP for the mobile release process (release-raccoon).
+ * @param {string} projectId The GCP project ID to send logs to.
+ * @param {string} logName The name of the log to send logs to in GCP.
+ * @param {string} logLevel The minimum severity level to log.
+ * @param {boolean} enableConsoleLogs If true, logs will be sent to the console as well as GCP.
+ * @param {[key: string]: string} labels K/V metadata that will be add to the logs
```
somewhatabstract, line 11: **suggestion:** Since there's only one parameter that is an object of options, you should create a type for that above the function (or in `types.ts`) and document each property in that type, then just use the type in a single `@param` here.

packages/wonder-stuff-ci/src/get-mobile-release-logger.ts
somewhatabstract: Yes, as you saw with @jeresig's suggestion, [here][1] (adding this comment as a paper trail for anyone reading this review later).

[1]: https://github.com/Khan/wonder-stuff/pull/636/files#r1164180337

packages/wonder-stuff-ci/src/get-mobile-release-logger.ts
somewhatabstract: The easiest way to handle that, is this, I believe.
```suggestion
     {
        projectId = "mobile-365917",
        logName = "release-raccoon",
        logLevel = "info",
        enableConsoleLogs = false,
        labels = {},
     } = {},
```

If that doesn't work then you could do:

```
myFunc(options?: MyOptionsType) {
    const {
        projectId = "mobile-365917",
        logName = "release-raccoon",
        logLevel = "info",
        enableConsoleLogs = false,
        labels = {},
    } = options ?? {};
```

packages/wonder-stuff-ci/src/get-mobile-release-logger.ts
somewhatabstract: It might be necessary to export the relevant pieces from Wonder Stuff Server if they aren't already, but yes, we should try to share what we can here. I think it's ok that this new package takes a dependency on WS Server.

packages/wonder-stuff-ci/src/__tests__/get-mobile-release-logger.test.ts
```diff
@@ -0,0 +1,12 @@
+import {getMobileReleaseLogger} from "../get-mobile-release-logger";
+
+describe("#getMobileReleaseLogger", () => {
+    it("return the mobile release logger", () => {
+        // Act
+        const result = getMobileReleaseLogger();
+
+        // Assert
+        expect(result).toBeTruthy();
+        expect(result.level).toStrictEqual("info");
```
jlauzy, line 10: Not sure what other tests I can add for these getter functions 🤔

packages/wonder-stuff-ci/src/get-mobile-release-logger.ts
jlauzy: I got it to work with the double fallback that John mentioned it looks like this.

```
export const getMobileReleaseLogger = (
    {
        redirectToStdout,
        labels,
    }: {
        redirectToStdout: boolean;
        labels: {[key: string]: string};
    } = {
        redirectToStdout: true,
        labels: {},
    },
): winston.Logger => {
```

packages/wonder-stuff-ci/src/get-mobile-release-logger.ts
jlauzy: ooo good point, i'll add that type!

jlauzy:
The parent pull-request ([#628](https://github.com/Khan/wonder-stuff/pull/628)) has been merged into `main`, but this branch (`FEI-5057-AddLoggerToGCPForCI`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

jlauzy:
Huh looks like one of the Flow type creation is broken? Any idea of how to resolve it?
image

jeresig:
@kevinbarabash Looks like some flow type generation logic may be broken, ideas on how to fix it?

kevinbarabash:
I think a quick fix might be to fix the type syntax in the comment.  There shouldn't be a `}` after the `]`.  I'll check out the PR and see if that fixes things.

kevinbarabash:
The syntax in the source code looks correct.  I'm not sure why the generate .d.ts file would end up with incorrect syntax.  I've checked out the PR and am investigating now.

kevinbarabash:
The issue is the original type in the jsdoc was `{[key: string]: string}`, but the first set of `{}` are stripped off which results in `[key: string]: string` which isn't a valid TypeScript type.  Adding a second set of `{}` fixed the issue.

jeresig (approved):
Great - thanks Jeffrey!

PR #639, opened 2023-04-13
Update 'imports-requiring-flow' eslint rule to support regex patterns

kevinbarabash:
## Summary:
We want to ensure all components using HOCs are using Flow in webapp.  If we detect any that aren't we'll either updating them to use Flow or check that their usages very careful.  In particular we want to make sure that consumers of components using HOCs aren't using ref forwarding.

Issue: None

## Test plan:
- cd packages/eslint-plugin-khan
- yarn test


kevinbarabash:
I'm not sure why these lint issues didn't come up in #638, but I'm going to fix them in this PR so we don't end up with even more lint.

package.json
```diff
@@ -87,7 +87,7 @@
     "coverage": "yarn run jest --coverage",
     "format": "prettier --write .",
     "lint": "yarn lint:ci .",
-    "lint:ci": "eslint --config .eslintrc.js --ext .ts --ext .js",
```
kevinbarabash, line 90: Using the `--config` option overrides the normal cascading behavior of eslintrc files.

packages/eslint-plugin-khan/.eslintrc.js
```diff
@@ -0,0 +1,11 @@
...
+module.exports = {
+    rules: {
+        "no-console": "off",
+        "import/extensions": "off",
+        "import/no-commonjs": "off",
+        "import/order": "off",
+        "@typescript-eslint/no-var-requires": "off",
+        "@typescript-eslint/no-unused-vars": "off",
+    },
+};
```
kevinbarabash, line 11: We can tackle these later.

packages/eslint-plugin-khan/lib/rules/jest-async-use-real-timers.js
```diff
@@ -5,7 +5,7 @@ const t = require("@babel/types");
  *
  * @param {CallExpression} describeCall the node for a call to `describe`.
  */
-const findBeforeEach = describeCall => {
+const findBeforeEach = (describeCall) => {
```
kevinbarabash, line 8: I ran `yarn lint --fix` and it fixed the prettier issues.

packages/eslint-plugin-khan/lib/rules/jest-await-async-matchers.js
```diff
@@ -1,16 +1,16 @@
 const t = require("@babel/types");
 
-const getPropertyNames = node => {
+const getPropertyNames = (node) => {
     if (!t.isMemberExpression(node)) {
-        [];
+        return [];
```
kevinbarabash, line 5: whoops

packages/wonder-stuff-ci/src/extract-mobile-release-info-from-branch-name.ts
```diff
@@ -1,3 +1,5 @@
+import type {MobileReleaseBranchInfo} from "./types";
```
kevinbarabash, line 1: I'm not sure why TypeScript didn't say anything about this missing import.

packages/wonder-stuff-render-server/types/superagent-cache-plugin.d.ts
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 import {SuperAgentRequest, Response} from "superagent";
```
kevinbarabash, line 2: The eslint rule thinks `SuperAgentRequest` isn't being used because it doesn't understand interface merging. 😞

packages/wonder-stuff-render-server/types/.eslintrc.js
```diff
@@ -1,6 +1,5 @@
 /* eslint-disable import/no-commonjs */
 module.exports = {
-    extends: ["../.eslintrc.js"],
```
kevinbarabash, line 3: This isn't necessary after getting rid of the `--config` option we were passing to `eslint`.

.github/workflows/node-ci.yml
```diff
@@ -83,7 +83,7 @@ jobs:
       uses: Khan/actions@full-or-limited-v0
       with:
         full-trigger: ${{ steps.eslint-reset.outputs.filtered }}
-        full: yarn lint:ci
+        full: yarn lint:ci .
```
kevinbarabash, line 86: Without the `.` eslint was only running on the .js files in the root directory.  I don't know why this is the default eslint behavior, it feels like such a footgun. 😞

kevinbarabash:
This was previously approved and all of the checks are passing now so even though there were some changes that affect other packages I'm going to go ahead and land this.

PR #646, opened 2023-04-14
Add 'jsdoc/valid-types' rule to our eslint config

kevinbarabash:
## Summary:
Apparently flowgen tries to convert types in comments from TS to Flow syntax.  It will fail to convert a file with invalid TS types in their docstrings.  This PR fixes existing issues and prevents future regressions.  I'm not sure why the errors it caught weren't causing flowgen to fail, but there was such an issue in #636.

Issue: None

## Test plan:
- yarn lint


jeresig (approved):
Ahh interesting - thank you!

PR #647, opened 2023-04-14
Fix sync-tag eslint rule

kevinbarabash:
## Summary:
The tests didn't catch this because they were mocking it.  The reason why we need to bother with these util wrappers around functions is that RuleTester doesn't provide a way to mock things so I ended up having to mock things in a hacky way.  I added an assert() call to the place where we do the mocking, but I think we need rethink how we write eslint tests.  We should probably create our own jest-based test harness, but that'll take some doing so I'm going to punt on it for now.

Issue: None

## Test plan:
- yarn --cwd packages/eslint-plugin-khan test

### DIFF
--- a/packages/eslint-plugin-khan/lib/util.js
+++ b/packages/eslint-plugin-khan/lib/util.js
@@ -1,6 +1,7 @@
-const {execFile} = require("child_process");
+const {execFile, execSync} = require("child_process");
 
 // This is done so that we can override execSync in the tests
 module.exports = {
     execFile,
+    execSync,
 };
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-khan/lib/util.js
```diff
@@ -1,6 +1,7 @@
-const {execFile} = require("child_process");
+const {execFile, execSync} = require("child_process");
```
jeresig, line 1: It seems like using types would've solved this specific issue, as well, right?

jeresig (approved):

packages/eslint-plugin-khan/lib/util.js
kevinbarabash: That too.  This code needs a lot of TLC.

PR #651, opened 2023-04-14
Set up typedoc to generate documentation from jsdoc comments

kevinbarabash:
## Summary:
[typedoc](https://typedoc.org/) can generate nice documentation sites from jsdoc comments.  This PR configures the tool to generate documentation.  It's themeable so if we don't like the theme we can swap it out for something else, see https://typedoc.org/guides/themes/.

TODO:
- [x] publish docs/ folder using the same gh-pages action as wonder-blocks

Issue: None

## Test plan:
- yarn build:docs
- open docs/index.html

Screen Shot 2023-04-14 at 5 04 56 PM
Screen Shot 2023-04-14 at 5 04 29 PM

The default theme supports dark mode out of the box:
Screen Shot 2023-04-14 at 5 31 18 PM


jeresig (approved):
Very cool - this is going to be really useful for folks!

PR #660, opened 2023-04-28
Move TS enum eslint rules into config package

kevinbarabash:
## Summary:
By moving these rules into the eslint-config-khan package, we can more easily adopt them in other repos.  In particular, if a repo is using dependabot then these new rules should be added automatically.

Issue: None

## Test plan:
- publish, see if dependenbot creates a PR in wonder-blocks with this change

### DIFF
--- a/packages/eslint-config-khan/index.js
+++ b/packages/eslint-config-khan/index.js
@@ -130,5 +130,7 @@ module.exports = {
             ERROR,
             {args: "none", varsIgnorePattern: "^_*$"},
         ],
+        "@typescript-eslint/prefer-enum-initializers": "error",
+        "@typescript-eslint/prefer-literal-enum-member": "error",
     },
 };
--- a/packages/eslint-config-khan/index.js
+++ b/packages/eslint-config-khan/index.js
@@ -89,7 +89,7 @@ module.exports = {
         /**
          * prettier rules
          */
-        "prettier/prettier": ["error", require("./.prettierrc")],
+        "prettier/prettier": [ERROR, require("./.prettierrc")],
 
         /**
          * react rules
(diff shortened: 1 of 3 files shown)


jeresig (approved):

packages/eslint-config-khan/index.js
```diff
@@ -130,5 +130,7 @@ module.exports = {
             ERROR,
             {args: "none", varsIgnorePattern: "^_*$"},
         ],
+        "@typescript-eslint/prefer-enum-initializers": "error",
+        "@typescript-eslint/prefer-literal-enum-member": "error",
```
kevinbarabash, line 134: @jeresig I'm going to add comments about these before landing the PR.

packages/eslint-config-khan/index.js
```diff
@@ -125,10 +125,37 @@ module.exports = {
...
-        "@typescript-eslint/no-explicit-any": "off",
+
+        // Disallow the `any` type.
+        //
+        // We've disabled this rule because most repos have too many uses of `any`
+        // for this to be useful.  Individual repos can enable this rule if they're
+        // in a position to do so.
+        //
+        // Docs: https://typescript-eslint.io/rules/no-explicit-any
+        "@typescript-eslint/no-explicit-any": OFF,
```
kevinbarabash, line 136: Even if the rule is self explanatory, I think that it's useful to have a rationale as to why we're enabling/disabling it.

packages/eslint-config-khan/index.js
```diff
@@ -125,10 +125,37 @@ module.exports = {
...
+        // Docs: https://typescript-eslint.io/rules/prefer-enum-initializers
+        "@typescript-eslint/prefer-enum-initializers": ERROR,
+
+        // Require all enum members to be literal values.
+        //
+        // Non-literal values can make it hard to tell what the value of a
+        // enum member is.
+        //
+        // Docs: https://typescript-eslint.io/rules/prefer-literal-enum-member
+        "@typescript-eslint/prefer-literal-enum-member": ERROR,
```
kevinbarabash, line 159: @jeremywiebe I'm adding documentation about the rules here.

jeremywiebe: Great idea. Sometimes the bare rule name isn't entirely obvious. Thanks.

packages/eslint-config-khan/index.js
```diff
@@ -125,10 +125,37 @@ module.exports = {
...
+        //
+        // By default enums are assigned integer values starting with 0
+        // and ascending.  This can cause issue when removing or inserting
+        // new members in the enum since this can change their values.
+        // Requiring members to be explicitly initialized prevents this issue.
+        //
+        // Docs: https://typescript-eslint.io/rules/prefer-enum-initializers
+        "@typescript-eslint/prefer-enum-initializers": ERROR,
+
+        // Require all enum members to be literal values.
```
jeremywiebe, line 153: I know this has landed and I know this is how the docs talk about this, but to me, "literal values" isn't immediately obvious. I wonder if it'd be useful to use the term "explicit values" somewhere in here?  What we're preventing with this rule is the construction of enums where the value is implicitly assigned by the TS compilation, right?

kevinbarabash: That's a good call out.  `prefer-enum-initializers` requires explicit values.  The other rule requires the explicit values to be literals, e.g. `"hello"`, `5`, etc. as opposed to identifiers or function calls.  I can provide some examples in the docs to clarify, although there are links to the official docs which already have very detailed examples.

PR #662, opened 2023-05-01
FEI-5116: Refactor tsconfig files to properly spearate typecheck and build configs

kevinbarabash:
## Summary:
TypeScript config files named tsconfig.json will cascade in a similar way to how .eslintrc.js files cascade.  The tsconfig.json files in package were meant to only be used with the tsconfig-build.json and not the tsconfig.json at the root.  As a result this was causing issues for other tooling we want to use such as @typescript-eslint rules requiring types and typescript-coverage-report.  See https://github.com/Khan/perseus/pull/503 for additional details.

I ran into a couple of issues in this PR. . In order to get our superagent .d.ts override working for both 'yarn typecheck' and 'yarn build:types', I needed to move the override into the file that uses it.  Hopefully we can work out how to make this override work in its own file again in the future.

There was also an issue with 'parse5' which is used by the @types/superagent.  Updating to TS 5.x fixed this issue.

Issue: FEI-5116

## Test plan:
- yarn clean
- yarn typecheck
- yarn build:types


jeresig (approved):

packages/wonder-stuff-render-server/src/requests-from-cache.ts
```diff
@@ -57,6 +57,31 @@ export const asUncachedRequest = (
...
+        pruneHeader(pruneHeader: Array<string>): this;
+        prune(
+            prune: (
+                response: SAResponse,
+                gutResponse: (response: SAResponse) => any,
+            ) => any,
+        ): this;
+        responseProp(responseProp: string): this;
+    }
+}
```
kevinbarabash, line 83: In Perseus we have .d.ts files in a types/ directory at the base.  That didn't work in this situation.  It must have something to do with the fact that this is extending a module's type whereas those are starting from scratch.  Another weird thing is that when this file was in a separate .d.ts file it worked with tsconfig-build.json but not tsconfig.json.  I can't explain that either. 😞

PR #696, opened 2023-05-11
5134.2: Create lint rule to prevent TS error suppressions on JSXAttributes

kevinbarabash:
## Summary:
Using @ts-expect-error comments on JSXAttributes, opts TS from checking all of the attrs/props on a JSX element.  This PR adds a lint rule to prevent this from happening by disallow TS error suppression comments on JSXAttributes.

Issue: FEI-5134

TODO
- [x] add comments in the code about what's going on with the comment/node ranges
- [x] try to make this a fixable rule by removing the offending comment from Program

## Test plan:
- yarn test ts-no-error-suppressions

### DIFF
--- a/packages/eslint-plugin-khan/src/rules/ts-no-error-suppressions.ts
+++ b/packages/eslint-plugin-khan/src/rules/ts-no-error-suppressions.ts
@@ -10,6 +10,16 @@ type MessageIds = "errorString";
 
 const message = "{{type}} not allowed on JSXAttributes";
 
+function rangesDoNotIntersect(
+    aRange: TSESTree.Range,
+    bRange: TSESTree.Range,
+): boolean {
+    const [aMin, aMax] = aRange;
+    const [bMin, bMax] = bRange;
+
+    return aMax < bMin || aMin > bMax;
+}
+
 export default createRule<Options, MessageIds>({
     name: "ts-no-error-suppressions",
     meta: {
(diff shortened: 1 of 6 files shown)


packages/eslint-plugin-khan/src/rules/ts-no-error-suppressions.ts
```diff
@@ -0,0 +1,91 @@
...
+
+const createRule = ESLintUtils.RuleCreator(
+    (name) =>
+        `https://github.com/Khan/wonder-stuff/blob/main/packages/eslint-plugin-khan/docs/${name}.md`,
+);
+
+type Options = ["always" | "never"];
+type MessageIds = "errorString";
+
+const message = "{{type}} not allowed on JSXAttributes";
```
drewkiimon, line 11: Interesting. It will have reference to the type in question?

kevinbarabash: The `{{type}}` gets replaced by the `data.type` provide by the `context.report({...})` call at the bottom of the file.  Unfortunately the docs aren't great so I've mainly been learning by looking at some of the rules in https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/eslint-plugin/src/rules.

kevinbarabash: Here's some snippets that were helpful in figuring this out:
- https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/explicit-member-accessibility.ts#L59-L65 
- https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/explicit-member-accessibility.ts#L298-L302

jeresig (approved):
Ahh - great! Thanks for adding this rule!

PR #697, opened 2023-05-11
FEI-5141.1: Add isTruthy and isNonNullable type predicates to wonder-stuff-core

kevinbarabash:
## Summary:
.filter(Boolean) doesn't work in TS like it does in Flow. To address this issue, I'm adding a couple of type predicates to wonder-stuff-core.  I'll be updating webapp to use these predicates before the migration.  I'll also be creating a fixable eslint rule for this so that people trying to use '.filter(Boolean)' after the migration can easily move to the new pattern.

Issue: FEI-5141

TODO:
- [x] have a custom flow override for type-predicates.ts so that these predicates work under Flow as well

## Test plan:
- yarn tsc
- yarn test type-predicates
- yarn build:docs
- open docs/index.html
- see the beautiful docs

Screen Shot 2023-05-12 at 2 27 06 PM
Screen Shot 2023-05-12 at 2 27 09 PM

### DIFF
--- a/packages/wonder-stuff-core/src/type-predicates.ts
+++ b/packages/wonder-stuff-core/src/type-predicates.ts
@@ -1,0 +1,29 @@
+type Falsy = false | 0 | "" | null | undefined;
+
+/**
+ * This is a type predicate - if `x` is "truthy", then it's `T`.
+ *
+ * This can be used with Array's `.filter()` method to remove non-truthy values
+ * from an array, e.g.
+ *
+ *   const array = [0, 5, false, true, "", "hello", undefined, null];
+ *   const truthyArray = array.filter(isTruthy);
+ *
+ * `truthyArray` will be `[5, true, "hello"]` and will have the following type:
+ * `Array<string | number | true>`.
+ */
+export const isTruthy = <T>(x: T | Falsy): x is T => !!x;
+
+/**
+ * This is a type predicate - if `x` is neither `null` or `undefined`, then it's `T`.
+ *
+ * This can be used with Array's `.filter()` method to remove non-truthy
+ * values from an array, e.g.
+ *
+ *   const array = [0, 5, false, true, "", "hello", undefined, null];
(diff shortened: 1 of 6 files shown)


packages/wonder-stuff-core/src/type-predicates.ts
```diff
@@ -0,0 +1,29 @@
...
+ * This can be used with Array's `.filter()` method to remove non-truthy
+ * values from an array, e.g.
+ *
+ *   const array = [0, 5, false, true, "", "hello", undefined, null];
+ *   const nonNullableArray = array.filter(isNonNullable);
+ *
+ * `nonNullableArray` will be `[0, 5, false, true, "", "hello"]` and will have the
+ * following type: `Array<string | number | boolean>`.
+ */
+export const isNonNullable = <T>(x: T | null | undefined): x is T => x != null;
```
jeremywiebe, line 29: I realize this is meant to be a type predicate, but to me, the naming would be a touch clearer if it was named `isNonNull`.   nullable feels like a type assertion whereas this truly is a predicate function.  `.filter(isNonNull)` also feels clearer than `.filter(isNonNullable)`  

Just some ramblings from an anonymous passerby. 🤷‍♂️  😄

jeresig: Hmm - that's a good point Jeremy, but it's tricky as this also filters out `undefined`, not just `null`... fwiw, TypeScript seems to use the terminology "isNonNullable" so it at least matches that?

jeresig (approved):
Thanks for adding these, Kevin!

packages/wonder-stuff-core/src/type-predicates.ts
kevinbarabash: I'll update the docs to call attention to the fact that this filters out `undefined` as well.  It might be useful to link to https://www.typescriptlang.org/docs/handbook/utility-types.html#nonnullabletype since that's what inspired the naming.

PR #701, opened 2023-05-12
Tweak eslint rule options to not use "always"/"never"

kevinbarabash:
## Summary:

We use "error", "warn", or "off" to control whether a rule runs or not.  The "always"/"never" option doesn't really make sense for the rules we currently have so I've removed those options from all of our eslint rules in this PR.

## Test Plan:
- yarn test eslint-plugin

### DIFF
--- a/.changeset/grumpy-hats-dress.md
+++ b/.changeset/grumpy-hats-dress.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/eslint-plugin": major
+---
+
+Remove always/never from eslint rule options
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ node_modules
 coverage
 dist
 docs
+!packages/eslint-plugin-khan/docs
 !.vscode
 .DS_Store
 yarn-error.log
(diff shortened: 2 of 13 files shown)


.changeset/grumpy-hats-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/eslint-plugin": major
+---
+
+Remove always/never from eslint rule options
```
kevinbarabash, line 5: @drewkiimon these files are created by running `yarn changeset add` and describe how the version of a particular package should be bumped when published.

.gitignore
```diff
@@ -2,6 +2,7 @@ node_modules
 coverage
 dist
 docs
+!packages/eslint-plugin-khan/docs
```
kevinbarabash, line 5: `docs` was prevent new files from being added in packages/eslint-plugin-khan/docs.

jeresig (approved):
Great! Thanks for making this a major release.

kevinbarabash:
A couple of checks failed because there's an issue with the `check-for-changeset` action.  I checked and this PR has a changeset file.  I also ran `node utils/pre-publish-check-ci.js` manually and it succeeded.

PR #702, opened 2023-05-12
FEI-5142: Add UnreachableCaseError for use in exhaustive switch-case statements

kevinbarabash:
## Summary:
In webapp and mobile we have some exhaustive switch-case statements using `(value: empty)` in the `default` case so that Flow can ensure that we've handled all of the possible cases.  TS doesn't have the same kind of type assertions so we'll use `UnreachableCaseError` instead.  The constructor for this error class takes a parameter that is unassignable (it's typed as `never`).  This means if we don't handle all of the cases, TS will complain because the `default` case will be reachable now.  This same approach will work with Flow as will.  Once I've published wonder-stuff-core, I'll update webapp to make use of it.  The mobile repo has already been migrated to TS, but can still make use of this new error class.

Issue: FEI-5142

## Test plan:
- `yarn test unreachable-case-error`
- `yarn build:docs`
- `yarn tsc`
- `open docs/index.html`, see the docs for `UnreachableCaseError`

Screen Shot 2023-05-12 at 5 39 15 PM

### DIFF
--- a/.changeset/sweet-apes-decide.md
+++ b/.changeset/sweet-apes-decide.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-stuff-core": minor
+---
+
+Add UnreachableCaseError for use in exhaustive switch-case statements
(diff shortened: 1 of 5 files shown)


jeresig (approved):
Nice - thanks for adding this!

PR #704, opened 2023-05-15
Update eslint-config-khan to include @khanacademy/ts-no-error-suppressions

kevinbarabash:
## Summary:
The ts-no-error-suppressions error should be used in all of our repos with React code that have been converted to TypeScript. This PR updates our eslint config to include this rule.

Issue: None

## Test plan:
- yarn build
- reload the eslint plugin in VSCode
- see that VSCode is no longer complaining about it not being able to find the rule
- node utils/pre-publish-check-ci.js  

NOTE: Khan/actions@check-for-changet will fail until https://github.com/Khan/actions/pull/44 is landed.

### DIFF
--- a/packages/eslint-config-khan/package.json
+++ b/packages/eslint-config-khan/package.json
@@ -9,16 +9,16 @@
     "repository": "git@github.com:Khan/eslint-config-khan.git",
     "author": "Kevin Barabash <kevinb@khanacademy.org>",
     "license": "MIT",
-    "dependencies": {},
     "peerDependencies": {
+        "@khanacademy/eslint-plugin": "^1.4.2",
         "@typescript-eslint/eslint-plugin": "^5.52.0",
         "@typescript-eslint/parser": "^5.52.0",
         "eslint": "^8.38.0",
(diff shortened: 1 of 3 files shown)


packages/eslint-config-khan/package.json
```diff
@@ -9,16 +9,16 @@
...
         "@typescript-eslint/eslint-plugin": "^5.52.0",
         "@typescript-eslint/parser": "^5.52.0",
         "eslint": "^8.38.0",
         "eslint-config-prettier": "^8.3.0",
+        "eslint-import-resolver-typescript": "^3.5.3",
         "eslint-plugin-import": "^2.23.4",
         "eslint-plugin-jsx-a11y": "^6.4.1",
         "eslint-plugin-prettier": "^3.4.0",
-        "eslint-plugin-react": "^7.24.0",
-        "eslint-import-resolver-typescript": "^3.5.3"
```
kevinbarabash, line 22: Using `yarn add ...` causes these imports to be reorder (sorted).

jeresig (approved):

PR #754, opened 2023-06-06
Add more configuration options for MobileReleaseLogger

jlauzy:
## Summary:
- Just adding more configurations options.

Issue: XXX-XXXX

## Test plan:


jeresig (approved):

somewhatabstract (changes requested):
Looks good; needs a changeset to be added before you can land this. I suspect this would be a `minor` update since it is adding new things but code using the old API would still work. Is that correct? If old code would break as soon as updating to this, then I think it would need to be a `major` update.

jlauzy:
> Looks good; needs a changeset to be added before you can land this. I suspect this would be a `minor` update since it is adding new things but code using the old API would still work. Is that correct? If old code would break as soon as updating to this, then I think it would need to be a `major` update.

@somewhatabstract added the changeset! Going to make a minor instead of major bump because no one was using the logger class since it was broken 😅

somewhatabstract (approved):

PR #759, opened 2023-06-07
[fixwscibuild] Make sure peer dependencies are added for wonder-stuff-ci

somewhatabstract:
## Summary:
This ensures that we don't try to inline the logging-winston code, which incorrectly excludes some of the code

Issue: XXX-XXXX

## Test plan:
`yarn build` and check that `./common` is not in the index.js for wonder-stuff-ci.

We'll also test this in @jlauzy's work where this issue was identified.


jeresig (approved):
Ahh - tricky, thank you!

PR #768, opened 2023-06-09
Cleaned up the code and fix the logger for new required type by GCP API

jlauzy:
## Summary:
- The logger no longer works because GCP's API made a breaking change where the resource type is required.
- Cleaned up the code to have more flexibility for these changes by exposing more configuration to the client.

image

I'm pretty sure they are going to make a change to their SDK so that this is no longer nullable but required
image

Issue: XXX-XXXX

## Test plan:
Works Again!
image

### DIFF
--- a/packages/wonder-stuff-ci/package.json
+++ b/packages/wonder-stuff-ci/package.json
@@ -6,7 +6,7 @@
         "node": ">=16"
     },
     "name": "@khanacademy/wonder-stuff-ci",
-    "version": "1.0.0",
+    "version": "1.1.0",
     "description": "Functions for automation and scripts.",
     "module": "dist/es/index.js",
     "main": "dist/index.js",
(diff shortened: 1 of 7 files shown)


packages/wonder-stuff-ci/src/get-mobile-release-logger.ts
```diff
@@ -7,50 +7,36 @@ import {
...
-        redirectToStdout: boolean;
-        labels: {[key: string]: string};
-        logLevel: GCPLogLevels;
-        logName: string;
-        projectId: string;
-    } = {
-        redirectToStdout: false,
-        labels: {},
-        logLevel: "info",
-        logName: "release-raccoon",
```
jlauzy, line 39: This looked very gnarly for a function signature, so I turned it into a type to make it more readable.

packages/wonder-stuff-ci/package.json
```diff
@@ -6,7 +6,7 @@
         "node": ">=16"
     },
     "name": "@khanacademy/wonder-stuff-ci",
-    "version": "1.1.0",
+    "version": "1.0.0",
```
jeresig: Err, that doesn't seem right! I think you have to revert this change - the version changes will come as part of the Changesets release process.

PR #802, opened 2023-06-28
Fix the path for the checksync binary

kevinbarabash:
## Summary:
The old code was using a relative directory path.  This worked when webapp/services/static was responsible for installing this plugin.  Now @khacademy/eslint-config installs this which results in @khanacademy/eslint-plugin ending up in a different location (it's nested inside @khanacademy/eslint-config's node_modules).

This PR fixes the issue by using ancesdir to find the nearest package.json.

## Test Plan:
- checkout fei-5181 in webapp
- re-enable the sync-tag rule in .eslintrc.ts.js and .eslintrc.flow.js
- use yarn link (or yarn pack) and replace webapp/services/static's version of @khanacademy/eslint-plugin with the one from this PR
- yarn eslint javascript/data-access/with-user-profile.jsx, see that it runs and reports an out of date sync tag

```
kevinbarabash@Kevins-MacBook-Pro static % yarn eslint javascript/data-access/with-user-profile.js
yarn run v1.22.19
$ eslint --ext .js --ext .jsx --quiet javascript/data-access/with-user-profile.js

... 4 other lint errors (these will be fixed in a separate PR)

  19:1  error  Looks like you changed the target content for sync-tag 'with-user-profile' in 'services/static/javascript/data-access/with-user-profile.tsx:18'. Make sure you've made the parallel changes in the source file, if necessary (1441908984 != 658500897)  @khanacademy/sync-tag

✖ 5 problems (5 errors, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```


packages/eslint-config-khan/package.json
```diff
@@ -9,6 +9,9 @@
     "repository": "git@github.com:Khan/eslint-config-khan.git",
     "author": "Kevin Barabash <kevinb@khanacademy.org>",
     "license": "MIT",
+    "dependencies": {
+        "ancesdir": "^5.0.0"
+    },
```
kevinbarabash, line 14: I meant to add this to `eslint-plugin-khan`. 🙃

kevinbarabash: It turns out `eslint-plugin-khan` already has `ancesdir` as a dep.

jeresig (approved):

PR #862, opened 2023-08-09
Ensure all packages have a repository url and directory

jeremywiebe:
## Summary:

While working in the Perseus repo I was trying to figure out something related to our `@khanacademy/eslint-config` settings. That package is published with a `repository` key pointing to an archived repo (https://github.com/Khan/eslint-plugin-khan). After some digging, I found that we'd moved the packages but not updated the repo urls. 

I've fixed them and enhanced our package.json linting to ensure we always have a repository url and directory (as per https://docs.npmjs.com/cli/v9/configuring-npm/package-json#repository). 

I've also moved these blocks up directly below the `description` key as it feels closely related (its package metadata.

Issue: "none"

## Test plan:

`yarn lint:pkg-json` passes

All repository data is correct


yarn.lock
jeremywiebe, line 1: Curious why yarn is making changes here.  I didn't change any dependencies, but it insists on these changes when I `yarn install`.

yarn.lock
```diff
@@ -4664,19 +4655,6 @@ flatted@^3.1.0:
   resolved "https://registry.yarnpkg.com/flatted/-/flatted-3.2.7.tgz#609f39207cb614b89d0765b477cb2d437fbf9787"
   integrity sha512-5nqDSxl8nn5BSNxyR3n4I6eDmbolI6WT+QqR547RwxQapgjQBmtktdP+HTBb/a/zLsbzERTONyUB5pefh5TtjQ==
 
-"flowgen@git+https://git@github.com/Khan/flowgen.git#fa8050929948fb4313f738301285fb44625678c4":
```
jeresig, line 4667: It looks like this wasn't cleaned up properly when this was removed? cc: @kevinbarabash

jeresig (approved):
Great! Thank you so much, very nice improvement.

yarn.lock
somewhatabstract: Looks like it's discovered some tidy up that can occur. :)

somewhatabstract (approved):
LGTM

Thanks for catching this and updating things.

somewhatabstract:
@jeremywiebe please add a changeset. It can be empty if you don't intend for a release to occur (just do `yarn changeset --empty`).

yarn.lock
somewhatabstract: I suspect it was removed from the `package.json` by hand but `yarn` wasn't re-run to update `yarn.lock`.

jeremywiebe:
> @jeremywiebe please add a changeset. It can be empty if you don't intend for a release to occur (just do `yarn changeset --empty`).

Thanks for catching that!  I'll do an empty one as these changes are fine to go out whenever the next release is cut.

PR #866, opened 2023-08-17
[fei5245.1.wsbuildsettings] Change build settings package to be in Khan namespace

somewhatabstract:
## Summary:
This renames the build settings package so that it is namespaced to Khan Academy. This ensures that someone cannot squat the package in NPM as they would need to be able to publish to the Khan Academy NPM org.

Issue: FEI-5245

## Test plan:
Not much to test here, but I did test that the package still builds.


jeremyspangler (approved):
Thanks for knocking this out!

PR #870, opened 2023-08-28
[aftereachrestoreenv] Add afterEachRestoreEnv

somewhatabstract:
## Summary:
Often, we want to modify environment variables during testing. However, this is a change that can easily bleed across test cases and so we must also properly handle resetting the variables after each test case. This is a common pattern that we can abstract away into a utility function, which I have done here.

Now, tests can just call `afterEachRestoreEnv` and it will automatically restore the environment variables to their original values after each test case.

Tests can be specific about the environment variables they want to restore, or restore the entire environment.

Issue: XXX-XXXX

## Test plan:
`yarn test`
`yarn typecheck`


packages/wonder-stuff-testing/src/jest/__tests__/after-each-restore-env.test.ts
```diff
@@ -0,0 +1,146 @@
...
+        } else {
+            process.env.ABSENT_1 = ABSENT_1;
+        }
+
+        if (ABSENT_2 === undefined) {
+            delete process.env.ABSENT_2;
+        } else {
+            process.env.ABSENT_2 = ABSENT_2;
+        }
+    });
```
somewhatabstract, line 41: **note:** This serves as a good example of why this is a useful thing to have. This whole thing could be replaced with:

```ts
afterEachRestoreEnv("EXPECT_1", "EXPECT_2", "ABSENT_1", "ABSENT_2");

// or, if we don't need to be exact:

afterEachRestoreEnv();
```

packages/wonder-stuff-testing/src/jest/after-each-restore-env.ts
```diff
@@ -0,0 +1,56 @@
...
+export const afterEachRestoreEnv = (
+    ...variableNames: ReadonlyArray<string>
+): void => {
+    // We capture the variables to restore. If none are given, we capture all.
+    const restoreAll = variableNames.length === 0;
+    const variablesToCapture = restoreAll
+        ? Object.keys(process.env)
+        : variableNames;
+    // We capture the current values on invocation.
+    const originalValues = variablesToCapture.reduce(
```
jeresig, line 20: Does this have to be in a `beforeEach`? So that the values match what they were before the test began to run - not when it was initialized?

jeresig (approved):
Oh nice! Yeah, this will be helpful, one question inline.

packages/wonder-stuff-testing/src/jest/after-each-restore-env.ts
somewhatabstract: I don't think so. In all the cases this is done in code, we snapshot the values at the start of the `describe`. The `beforeEach` would capture potentially mutated state and therefore allow for bleeding of changes across test cases, which we don't want.

packages/wonder-stuff-testing/src/jest/after-each-restore-env.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    // Now, in the afterEach call, we restore the environment state.
+    afterEach(() => {
+        // If we are restoriing all variables then we cannot rely solely on
+        // what was captured, but instead we must also check what is currently
+        // in the environment that was not captured to make sure we delete it.
+        if (restoreAll) {
+            for (const variableName of Object.keys(process.env)) {
+                if (!(variableName in originalValues)) {
+                    delete process.env[variableName];
```
somewhatabstract, line 47: **nit:** We could call `restoreValue(variableName, undefined)` here, just to be consistent.

PR #877, opened 2023-09-03
[updateprettierandrollup] Update prettier and rollup

somewhatabstract:
## Summary:
This updates Wonder Stuff to pretter v3 and rollup v3.

There's no functional changes here.

Issue: XXX-XXXX

### Work Sustainably:
I know it's a weekend. I just updated my personal projects for these two things. It seemed prudent to just apply what I learned while it was fresh rather than wait for a work day.

## Test plan:
`yarn lint`
`yarn build`


package.json
```diff
@@ -25,47 +25,47 @@
...
+    "@types/jest": "^29.5.4",
     "@types/jest-when": "^3.5.2",
     "@types/jsdom": "^21.1.1",
     "@types/superagent": "^4.1.16",
     "@types/winston": "^2.4.4",
-    "@typescript-eslint/eslint-plugin": "^5.62.0",
-    "@typescript-eslint/parser": "^5.62.0",
-    "babel-jest": "29.6.4",
+    "@typescript-eslint/eslint-plugin": "^6.5.0",
+    "@typescript-eslint/parser": "^6.5.0",
```
kevinbarabash, line 41: Cool!

package.json
```diff
@@ -25,47 +25,47 @@
...
+    "babel-jest": "^29.6.4",
     "babel-watch": "^7.8.1",
     "eslint": "^8.48.0",
-    "eslint-config-prettier": "^8.8.0",
-    "eslint-import-resolver-typescript": "^3.5.5",
+    "eslint-config-prettier": "^9.0.0",
```
kevinbarabash, line 45: FYI I ran into issues updating this in webapp.  I think it conflicted with dprint or something like that.

kevinbarabash: We don't use dprint here and it's only a dev dependency so there shouldn't be any issues making the upgrade in this repo.

package.json
```diff
@@ -79,11 +79,11 @@
...
   },
   "scripts": {
-    "build": "rollup -c build-settings/rollup.config.js",
-    "build:prodsizecheck": "rollup -c build-settings/rollup.config.js --configPlatforms='browser' --configFormats='esm' --configEnvironment='production'",
+    "build": "rollup -c build-settings/rollup.config.mjs",
+    "build:prodsizecheck": "yarn rollup --configPlatforms='browser' --configFormats='esm' --configEnvironment='production'",
     "build:types": "yarn tsc --build --verbose tsconfig-build.json",
     "build:docs": "typedoc",
-    "watch": "rollup -c build-settings/rollup.config.js --watch",
+    "watch": "yarn rollup --watch",
```
kevinbarabash, line 86: Why don't we need to pass `rollup` the location of the config file?

kevinbarabash (approved):
Thanks for updating these tools.  I think there's maybe an issue with some of the changes to the package.json scripts.  Please verify before landing.

package.json
somewhatabstract: Because it's running the `rollup` yarn script that already passes it.

kevinbarabash: I don't see a script called "rollup".

somewhatabstract: 🤦🏻 Yup, I never pushed that commit - it's still called `build` in this version. Doh! Thanks for catching that.

somewhatabstract: Pushed.

kevinbarabash (approved):
🎉

PR #905, opened 2023-09-18
Add moar helper types to wonder-stuff-core

kevinbarabash:
## Summary:
A number of repos are defining the same or similar utility types.  Let's bring them into wonder-stuff-core so that they're consistent across projects.

## Test Plan:
N/A

### DIFF
--- a/packages/wonder-stuff-core/src/types.ts
+++ b/packages/wonder-stuff-core/src/types.ts
@@ -37,3 +37,28 @@ export type Secrets = {readonly [key: string]: SecretString};
  * Make a read-only type mutable.
  */
 export type Mutable<T> = {-readonly [P in keyof T]: T[P]};
+
+/**
+ * Remove the keys in `R` from `C`.
+ */
+export type Without<C, R> = Omit<C, keyof R>;
+
+/**
+ * Opposite of `NonNullable<T>`.
+ *
+ * This type is most useful in making code that's been converted from Flow
+ * to TypeScript more readable.
+ */
+export type Nullable<T> = T | null | undefined;
+
+/**
+ * This type can be used in place of `any` to indicate that we don't know
+ * what the type is, but we know that it is wrong and needs to be fixed.
+ *
+ * It's similar to $FlowFixMe in Flow.
+ *
+ * This should be used in situations where `@ts-expect-error` cannot be used
+ * or should not be used (as is the case with type errors on props in JSX
+ * elements).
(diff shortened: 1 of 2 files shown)


packages/wonder-stuff-core/src/types.ts
```diff
@@ -37,3 +37,28 @@ export type Secrets = {readonly [key: string]: SecretString};
  * Make a read-only type mutable.
  */
 export type Mutable<T> = {-readonly [P in keyof T]: T[P]};
+
+/**
+ * Remove the keys in `R` from `C`.
+ */
+export type Without<C, R> = Omit<C, keyof R>;
```
kevinbarabash, line 44: I'm kind of questioning the usefulness of this type.  It's used primarily in components that use HOCs in webapp.  We only had a helper type before because Flow didn't have `Omit`.  I kind of think we should just replace `Without` with `Omit`.

lillialexis (approved):
Looks good though I don't know if I can weigh in on `Omit`/`Without`

packages/wonder-stuff-core/src/types.ts
somewhatabstract: I don't mind either way. It's a nice, clear utility that drops the need for the extra `keyof` and saves 2 keystrokes. However, if we didn't have it, I wouldn't mind either - but then you have to go find replace all the uses. Is that worth it? I don't know.

I don't think it does any harm existing.

packages/wonder-stuff-core/src/types.ts
```diff
@@ -37,3 +37,28 @@ export type Secrets = {readonly [key: string]: SecretString};
  * Make a read-only type mutable.
  */
 export type Mutable<T> = {-readonly [P in keyof T]: T[P]};
+
+/**
+ * Remove the keys in `R` from `C`.
```
somewhatabstract, line 42: **suggestion:** This feels a little inexact - perhaps:
```suggestion
 * Remove from `C` properties that have identical keys to those in `R`.
```

packages/wonder-stuff-core/src/types.ts
somewhatabstract, line 1: **suggestion[change-requested]:** Can you add types tests to verify these types are working as expected?

somewhatabstract (changes requested):
LGTM

Pushing back for changes so we can have some tests to verify these utility types work.

PR #990, opened 2023-12-14
[fei5402.1.node20workflows] Change workflows to node 20

somewhatabstract:
## Summary:
This updates our Wonder Blocks workflows to use Node 20.

In order to verify things, I learned some additional steps for macOS.

1. Switch to node 20 (I use NVM, so `nvm use 20`)
1. Run `yarn` or `yarn install`
  1. If you get an error about `node-gyp`. You either need to install XCode command line tools (You can do this by running `xcode-select --install` and following the prompts), or switch to use those tools from your version of XCode (XCode 14.3.1 appears to cause a lot of errors) - you can do this by running `sudo xcode-select -s /Library/Developer/CommandLineTools`.
  1. Re-run `yarn` or `yarn install` after this and all should be well.

Issue: FEI-5402

## Test plan:
See above steps to get things installed properly for macOS.

I then ran:
- `yarn test`
- `yarn lint`
- `yarn typecheck`
- `yarn build`


jeresig (approved):
It seems like we're getting the node-gyp error here, as well, so I think this will require a more-systemic fix.

somewhatabstract:
> It seems like we're getting the node-gyp error here, as well, so I think this will require a more-systemic fix.

There's a different issue for Linux than macOS. It'll need a different approach

somewhatabstract:
> It seems like we're getting the node-gyp error here, as well, so I think this will require a more-systemic fix.

Ah this is a Python 3.12 issue where the distutils module is deprecated. This won't affect devs right now since we're using 3.11.

I'll work on a "fix" for workflows. Short term would be to use 3.11.

PR #1018, opened 2024-02-15
[fixusageoutsidejest] Allow import outside of jest

somewhatabstract:
## Summary:
The `wait` function copied from Webapp verified that jest was in use at import. This worked when in webapp and it was just a file being imported, but now it's in Wonder Stuff Testing, it causes an error if any of the non-jest code is used outside of a jest environment, which is not what we want at all.

So, this removes that module level check, which in turn means Wonder Stuff Testing can be used in non-jest contexts, like stories.

See [Slack thread](https://khanacademy.slack.com/archives/CEA6W0F6F/p1707955461083099?thread_ts=1707867450.595019&cid=CEA6W0F6F) for more context, if you have access.

Issue: XXX-XXXX

## Test plan:
`yarn test`

Once published, we can update webapp and verify `dataFactoryFor` is usable within stories.

### DIFF
--- a/packages/wonder-stuff-testing/src/jest/wait.ts
+++ b/packages/wonder-stuff-testing/src/jest/wait.ts
@@ -1,8 +1,5 @@
 import {verifyRealTimers} from "./internal/verify-real-timers";
 import {unverifiedWait} from "./internal/unverified-wait";
-import {assertJest} from "./internal/assert-jest";
-
-assertJest();
 
 type WaitOptions = {
     /**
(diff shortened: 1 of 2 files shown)


packages/wonder-stuff-testing/src/jest/wait.ts
```diff
@@ -1,8 +1,5 @@
 import {verifyRealTimers} from "./internal/verify-real-timers";
 import {unverifiedWait} from "./internal/unverified-wait";
-import {assertJest} from "./internal/assert-jest";
-
-assertJest();
```
somewhatabstract, line 5: **note:** The `wait` call invokes a `verifyRealTimers` call, which in turn will call `assertJest` so we're still covered, just not at import time.

jeresig (approved):
Ah, nice!

evanchaney:
👏 

Thank you for addressing this.

PR #1021, opened 2024-04-05
[fei5484.1.node20actions] Update actions to Node 20 variants

somewhatabstract:
## Summary:
This updates our actions to use the Node 20 variants. This should mean almost no more Node 16 deprecation warnings, however, the `preactjs/compressed-size-action` has no Node 20 variant yet.

Issue: FEI-5484

## Test plan:
Put the PR up and see that the workflows run without any Node 16 deprecation warnings - except for those about `preactjs/compressed-size-action`.


jeresig (approved):

jeresig:
@jeremywiebe Were you able to solve the Node 16 error for the compressed-size-action in Perseus?

somewhatabstract:
> @jeremywiebe Were you able to solve the Node 16 error for the compressed-size-action in Perseus?

Regarding this. The repo updated to Node 16 a year ago, but never published a new version. So technically, it's specifying node 12 and GitHub is forcing it to run with Node 16. I would envisage this action is abandoned and we want to plan accordingly (we use it in Perseus, Wonder Stuff, and Wonder Blocks).

jeremywiebe:
@jeresig @somewhatabstract I don't remember seeing an error with the `compressed-size-action`, but it's a bit since I worked on this. 

I wonder if we should think about forking this action and just maintaining it ourselves in https://github.com/khan/actions ?  Or look at different solutions like:

  * https://github.com/bundlewatch/bundlewatch
  * https://github.com/LironEr/bundlemon (open source and looks pretty good!)
  * https://github.com/marketplace/actions/size-limit-action

PR #1024, opened 2024-08-29
[fei5815.2.workaround] Add step to tidy up files

somewhatabstract:
## Summary:
It seems that `changesets publish` isn't honoring our root `.npmignore` even though `npm pack` would. This works around that by borrowing some code from Wonder Blocks where we tidy up any ignored files that aren't source-controlled, reducing what we package.

We also explicitly include things in each package.json.

This combination means we can easily exclude files and folders in the `dist` location, without having really explicit `files` definitions that might be hard to maintain. 

Issue: FEI-5815

## Test plan:
`yarn build:types` and see the files being removed.


jeresig:
What happens if we put a symlink to the .npmignore (or a copy of it) in each of the package directories? That would make sense that it wouldn't work for a global .npmignore and instead would be per-package. If that could work I'd definitely prefer that to a custom script.

somewhatabstract:
> What happens if we put a symlink to the .npmignore (or a copy of it) in each of the package directories? That would make sense that it wouldn't work for a global .npmignore and instead would be per-package. If that could work I'd definitely prefer that to a custom script.

We already have the script - I just adapted it to use the `.npmignore` rather than hard-code all the things it deleted.

The issue with a symlinked `.npmignore` is getting the paths right. I think for that, we'd want the master file to not be a root `.npmignore` but rather some file like `npm-ignore-template` and then symlink that, so we're clear that the paths are not relative to root but relative to a package root.

I can give that a try. If that works then we can drop the script entirely rather than just edit it, then keep the `files` change and the type definitions check I added so that types are correct. This should reduce our package sizes significantly without breaking anything.

somewhatabstract:
OK, I tried the symlinks in Wonder Blocks too (which does a publish of pre-release packages, making this easier to test) and it completely fails.

```
🦋  error npm error path /home/runner/work/wonder-blocks/wonder-blocks/packages/wonder-blocks-toolbar/.npmignore
```

However, I think just having a copy of `.npmignore` in each folder and a script to overwrite with the template will provide an easy to update workaround. We could run the script manually when adding a package and just have a simple consistency test to catch when a package doesn't ahve one.

jeresig (approved):

PR #1082, opened 2024-11-28
Update TypeScript to 5.7.2

kevinb-khan:
## Summary:
Wonder Blocks is updating Storybook to 8.x which requires using "moduleResolution": "bundler" in its tsconfig, see https://stackoverflow.com/questions/72193784/why-doesnt-the-exports-field-of-npm-work-in-typescript for details.  To fix this, I decided to update TypeScript in Wonder Blocks.  Instead of updating to the minimally required version I figured I may as well upgrade to the most recent version since I'm upgrading anyways.

Although we probably don't need to update TypeScript in all of our repos at the same time (unless we start using new syntax), we may as well.  That way, if a new Wonder Blocks package is published with new TS syntax it its types, Perseus will be ready.

Issue: None

## Test plan:
- yarn typecheck

### DIFF
--- a/tsconfig-common.json
+++ b/tsconfig-common.json
@@ -26,6 +26,6 @@
 
         /* Completeness */
         "skipDefaultLibCheck": true, // it's safe to assume that built-in .d.ts files are correct
-        "skipLibCheck": false
+        "skipLibCheck": true, // there are conflicts between some of the globals in various third-party packages
     }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -1514,7 +1514,12 @@
   dependencies:
     eslint-visitor-keys "^3.3.0"
 
-"@eslint-community/regexpp@^4.5.1", "@eslint-community/regexpp@^4.6.1":
+"@eslint-community/regexpp@^4.10.0":
+  version "4.12.1"
+  resolved "https://registry.yarnpkg.com/@eslint-community/regexpp/-/regexpp-4.12.1.tgz#cfc6cffe39df390a3841cde2abccf92eaa7ae0e0"
+  integrity sha512-CCZCDJuduB9OUkFkY2IgppNZMi2lBQgD2qzwXkEia16cge2pijY/aXi96CJMquDMn3nJdlPV1A5KrJEXwfLNzQ==
+
+"@eslint-community/regexpp@^4.6.1":
   version "4.8.0"
   resolved "https://registry.yarnpkg.com/@eslint-community/regexpp/-/regexpp-4.8.0.tgz#11195513186f68d42fbf449f9a7136b2c0c92005"
   integrity sha512-JylOEEzDiOryeUnFbQz+oViCXS0KsvR1mvHkoMiu5+UiBvy+RYX7tzlIIIEstF/gVa2tj9AQXk3dgnxv6KxhFg==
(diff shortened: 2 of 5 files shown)


tsconfig-common.json
```diff
@@ -26,6 +26,6 @@
 
         /* Completeness */
         "skipDefaultLibCheck": true, // it's safe to assume that built-in .d.ts files are correct
-        "skipLibCheck": false
+        "skipLibCheck": true, // there are conflicts between some of the globals in various third-party packages
```
somewhatabstract, line 29: **question[change-required]:** Can you be more specific here? I think we should document which packages and which conflicts somewhere (either here or in a linked Jira ticket) so that we know when this check might be changed to `false` again.

yarn.lock
somewhatabstract, line 1: **question[no-change-required]:** I wonder if we should have `yarn.lock` marked as an auto-generated file so that it's not reviewed by default. What do you think?

somewhatabstract (changes requested):
Thanks for doing this. The changes look good to me except for one. I have left a question/suggestion inline; but to summarize, I think we need to be more explicit about which modules/globals conflict so that future us can properly understand the implications of future updates and know when certain config values may no longer be relevant.

tsconfig-common.json
kevinb-khan: Good idea, will do.

yarn.lock
kevinb-khan: I think that's a great idea.

tsconfig-common.json
```diff
@@ -26,6 +26,23 @@
 
         /* Completeness */
         "skipDefaultLibCheck": true, // it's safe to assume that built-in .d.ts files are correct
-        "skipLibCheck": false
+        // Setting `skipLibCheck: true`, avoids the following issues in various .d.ts files:
+        // - node_modules/@google-cloud/kms/node_modules/long/index.d.ts:446:1 - error
+        //   TS1203: Export assignment cannot be used when targeting ECMAScript modules.
+        //   Consider using 'export default' or another module format instead.
```
kevinb-khan, line 32: This one is not a conflict.  Maybe the .d.ts files will be different in a more recent version.  I don't want to make changes that could modify the runtime behavior.

tsconfig-common.json
```diff
@@ -26,6 +26,23 @@
...
+        // - node_modules/@typescript-eslint/utils/node_modules/@typescript-eslint/typescript-estree/dist/ts-estree/ts-nodes.d.ts:19:3551 - error
+        //   TS2694: Namespace 'ts' has no exported member 'InputFiles'.
+        // - node_modules/@typescript-eslint/utils/node_modules/@typescript-eslint/typescript-estree/dist/ts-estree/ts-nodes.d.ts:19:3567 - error
+        //   TS2694: Namespace 'ts' has no exported member 'UnparsedSource'.
```
kevinb-khan, line 36: `@typescript-eslint/typescript-estree` provides ambient type definitions that conflict with the TypeScript types for TypeScript itself.

.gitattributes
```diff
@@ -0,0 +1 @@
+yarn.lock binary
```
kevinb-khan, line 1: I though this would do the trick (this is what mobile does), but maybe I need to do:
```
yarn.lock linguist-generated
```
This is what webapp does.

.gitattributes
somewhatabstract, line 1: **praise:** Great! Thanks.

tsconfig-common.json
somewhatabstract: **praise:** Thanks. This is really helpful. It won't catch any new issues that come up, but it gives us something to understand.

somewhatabstract (approved):
LGTM

Thank you!

PR #1085, opened 2024-12-03
[FEI-6012]: Update eslint-{config|plugin}-khan to work with TypeScript v5.7.x

kevinb-khan:
## Summary:
This PR updates all `@typescript-eslint/*` packages to 8.17.0.  Earlier versions of `@typescript-eslint/typescript-estree` would print a warning message when loaded.  This necessitated updating `eslint-jest-plugin` which uses `@typescript-eslint/utils` which uses `@typescript-eslint/typescript-estree` internally.

I also needed to update eslint-plugin-khan's test harness to use `@typescript-eslint/rule-tester` since the old harness broke after the other upgrades.  The plugins themselves needed some tweaking after some of the types changed when updating `@typescript-eslint/utils`.

Issue: FEI-6012

## Test plan:
- yarn lint

PR #1090, opened 2024-12-09
Fix eslint-plugin-khan tests so that they can be run using jest

kevinb-khan:
## Summary:
#1085 caused an issue where tests were only running for one of the eslint plugin rules.  Thanks @beaesguerra for noticing this.  This PR updates our testing setup so that we can use jest to run the eslint-plugin-khan tests instead of mocha.  It also fixes a few failures that are due to changes in eslint and RuleTester's behaviour.

Issue: None

## Test plan:
- yarn test


jeresig (approved):
Ahh nice - thanks for the fixes!

kevinb-khan:
> Ahh nice - thanks for the fixes!

I love that we're able to use `RuleTester` with `jest` now.  I'll make the same change to services/static/dev/eslint-plugin-static-service at some point so that we can run those tests using GitHub Actions.

beaesguerra:
Thanks for the fix @kevinb-khan ! 🎉

PR #1091, opened 2024-12-11
Add aphrodite-add-style-variable-name lint rule

beaesguerra:
## Summary:

Adding a custom lint rule to ensure consistent naming for when we use `addStyle` for aphrodite styling. Variable naming should be in the format `StyledTag`.

By having predictable names for styled elements, we are able to use [component mapping](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y?tab=readme-ov-file#component-mapping) more effectively for `eslint-plugin-jsx-a11y` so it can use the underlying HTML for accessibility checks. 

Issue: FEI-5952
Related ADR: [ADR-781](https://khanacademy.atlassian.net/wiki/x/IoBVyg)

## Test plan:
- Tests pass
- Linting works in the demo project (will need some help with this!)

[ADR-781]: https://khanacademy.atlassian.net/browse/ADR-781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/eslint-plugin-khan/demo/.eslintrc.js
+++ b/packages/eslint-plugin-khan/demo/.eslintrc.js
@@ -22,5 +22,6 @@ module.exports = {
                 rootDir: __dirname,
             },
         ],
+        "@khanacademy/aphrodite-add-style-variable-name": "error",
     },
 };
--- a/packages/eslint-plugin-khan/demo/.eslintrc.js
+++ b/packages/eslint-plugin-khan/demo/.eslintrc.js
@@ -1,6 +1,7 @@
-/* eslint-disable import/no-commonjs */
 module.exports = {
+    root: true,
     plugins: ["@khanacademy"],
+    parser: "@typescript-eslint/parser",
     rules: {
         "@khanacademy/jest-await-async-matchers": [
             "error",
@@ -22,6 +23,7 @@ module.exports = {
                 rootDir: __dirname,
             },
         ],
-        "@khanacademy/aphrodite-add-style-variable-name": "error",
+        // TODO(kevinb): re-enable after publishing @khanacademy/eslint-plugin
+        // "@khanacademy/aphrodite-add-style-variable-name": "error",
     },
 };
--- a/packages/eslint-plugin-khan/demo/src/foo.tsx
+++ b/packages/eslint-plugin-khan/demo/src/foo.tsx
@@ -1,5 +1,11 @@
 import * as React from "react";
 
+function addStyle(str: string) {
+    // Placeholder example for calling addStyle for aphrodite
+}
+
+const div = addStyle("div");
+
 export const Icon = (): React.ReactNode => {
     return (
         <svg>
--- a/packages/eslint-plugin-khan/docs/aphrodite-add-style-variable-name.md
+++ b/packages/eslint-plugin-khan/docs/aphrodite-add-style-variable-name.md
@@ -1,0 +1,35 @@
+# Naming convention for addStyle variable (aphrodite-add-style-variable-name)
+
+The variable name when using `addStyle` should be the same as the tag argument in the format `StyledTag`.
+
+This is useful so that Aphrodite styled elements can be mapped to HTML elements for static code analysis. For example, if the `addStyle` variables are consistently named, we are able to provide custom component mapping to `eslint-plugin-jsx-a11y` so that it can identify linting issues based on the underlying HTML tag.
+
+## Rule Details
+
+The following are considered warnings:
+
+```ts
+const div = addStyle("div");
+```
+
+```ts
+const foo = addStyle("span");
+```
+
+```ts
+const container = addStyle("div");
+```
+
+The following are not considered warnings:
+
+```ts
+const StyledDiv = addStyle("div");
+```
+
+```ts
+const StyledSpan = addStyle("span");
+```
+
+```ts
+const StyledImg = addStyle("img");
+```
--- a/packages/eslint-plugin-khan/docs/aphrodite-add-style-variable-name.md
+++ b/packages/eslint-plugin-khan/docs/aphrodite-add-style-variable-name.md
@@ -1,8 +1,10 @@
 # Naming convention for addStyle variable (aphrodite-add-style-variable-name)
 
-The variable name when using `addStyle` should be the same as the tag argument in the format `StyledTag`.
+The variable name should match the tag name passed into `addStyle` and follow the format: `StyledTag`.
 
 This is useful so that Aphrodite styled elements can be mapped to HTML elements for static code analysis. For example, if the `addStyle` variables are consistently named, we are able to provide [custom component mapping](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y#component-mapping) to `eslint-plugin-jsx-a11y` so that it can identify linting issues based on the underlying HTML tag.
+
+This naming convention only applies to string arguments for HTML tags. It is not enforced when a component is used as the addStyle argument.
 
 ## Rule Details
 
@@ -32,4 +34,10 @@ const StyledSpan = addStyle("span");
 
 ```ts
 const StyledImg = addStyle("img");
+```
+
+```ts
+// Any variable name can be used when addStyle is used with a component
+const LinkWithStyle = addStyle(Link);
+const Foo = addStyle(Bar);
 ```
--- a/packages/eslint-plugin-khan/docs/aphrodite-add-style-variable-name.md
+++ b/packages/eslint-plugin-khan/docs/aphrodite-add-style-variable-name.md
@@ -2,7 +2,7 @@
 
 The variable name when using `addStyle` should be the same as the tag argument in the format `StyledTag`.
 
-This is useful so that Aphrodite styled elements can be mapped to HTML elements for static code analysis. For example, if the `addStyle` variables are consistently named, we are able to provide custom component mapping to `eslint-plugin-jsx-a11y` so that it can identify linting issues based on the underlying HTML tag.
+This is useful so that Aphrodite styled elements can be mapped to HTML elements for static code analysis. For example, if the `addStyle` variables are consistently named, we are able to provide [custom component mapping](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y#component-mapping) to `eslint-plugin-jsx-a11y` so that it can identify linting issues based on the underlying HTML tag.
 
 ## Rule Details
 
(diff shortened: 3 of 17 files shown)


packages/eslint-plugin-khan/demo/src/foo.tsx
```diff
@@ -1,5 +1,11 @@
 import * as React from "react";
 
+function addStyle(str: string) {
+    // Placeholder example for calling addStyle for aphrodite
+}
+
+const div = addStyle("div");
```
beaesguerra, line 7: I updated the demo project to include a linting error intentionally. Looks like it's [working in CI](https://github.com/Khan/wonder-stuff/actions/runs/12267532496/job/34227812456?pr=1091)! Some thoughts/questions:
- I'm not sure why it shows the errors twice in the PR (same with the other errors)
- Should CI be running linting in the demo project since we expect it to fail?
- I followed the steps in the [Demo readme](https://github.com/Khan/wonder-stuff/blob/main/packages/eslint-plugin-khan/demo/README.md) locally and was running into an error. The linting error also wasn't being shown in VSCode. The error:
```
Oops! Something went wrong! :(

ESLint: 8.40.0

TypeError: Failed to load plugin '@khanacademy' declared in '.eslintrc.js': Class extends value undefined is not a constructor or null

```

kevinb-khan: Try updating `eslint` to 8.55.0.

packages/eslint-plugin-khan/src/rules/aphrodite-add-style-variable-name.ts
```diff
@@ -0,0 +1,73 @@
...
+    defaultOptions: [],
+    create(context) {
+        return {
+            VariableDeclarator(node: TSESTree.VariableDeclarator) {
+                // Check if addStyle is being called
+                if (
+                    node.init &&
+                    node.init.type === "CallExpression" &&
+                    node.init.callee.type === "Identifier" &&
+                    node.init.callee.name === "addStyle"
```
kevinb-khan, line 38: style suggestion:  some of our other code compares `.type` fields to things like `AST_NODE_TYPES.Identifier`.  `AST_NODE_TYPES` can be imported from `"@typescript-eslint/utils"`.

packages/eslint-plugin-khan/src/rules/aphrodite-add-style-variable-name.ts
```diff
@@ -0,0 +1,73 @@
...
+                ) {
+                    // Get variable name for the addStyle return value
+                    const variableName =
+                        node.id.type === "Identifier" ? node.id.name : null;
```
kevinb-khan, line 42: style suggestion: this check could be an early return

packages/eslint-plugin-khan/src/rules/aphrodite-add-style-variable-name.ts
```diff
@@ -0,0 +1,73 @@
...
+
+                    // Get the tag name that was passed into addStyle
+                    const firstArg = node.init.arguments[0];
+                    if (
+                        firstArg &&
+                        firstArg.type === "Literal" &&
+                        typeof firstArg.value === "string"
+                    ) {
```
kevinb-khan, line 50: style suggestion: this check could be an early return

packages/eslint-plugin-khan/src/rules/aphrodite-add-style-variable-name.ts
```diff
@@ -0,0 +1,73 @@
...
+                        const tagName = firstArg.value;
+                        const expectedName = `Styled${tagName
+                            .charAt(0)
+                            .toUpperCase()}${tagName.slice(1)}`;
```
kevinb-khan, line 54: style nit: you may want to extract the part that does the capitalization into a helper function call `capitalizeString`.  This will make it easier for people to grok what's going on here.

packages/eslint-plugin-khan/src/rules/aphrodite-add-style-variable-name.ts
```diff
@@ -0,0 +1,73 @@
...
+
+                        // Check if the variable name matches the expected pattern
+                        if (variableName !== expectedName) {
+                            context.report({
+                                node: node.id,
+                                messageId: "errorString",
+                                data: {
+                                    variableName,
+                                    tagName,
+                                    expectedName,
```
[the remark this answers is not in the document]
kevinb-khan: It's probably good enough for now to ship this without the autofix.  `addStyle("` appears 127 times across 91 files.  Some of those will be invalid cases with this rule but fixing them by hand shouldn't take too long.  If you want to add autofixing in the future let me know and we can pair on it.

packages/eslint-plugin-khan/src/rules/aphrodite-add-style-variable-name.ts
kevinb-khan: style suggestion: this could be one or early returns to avoid excessive indentation, e.g.
```ts
if (node?.init.type !== "CallExpression") {
  return;
}
if (node.init.callee.type !== "Identifier" || node.init.callee.name !== "addStyle) {
  return;
}
...
```

packages/eslint-plugin-khan/test/rules/aphrodite-add-style-variable-name.test.ts
```diff
@@ -0,0 +1,120 @@
...
+        },
+        {
+            code: `const StyledButton = addStyle("button")`,
+        },
+        {
+            code: `const StyledP = addStyle("p")`,
+        },
+        {
+            code: `const StyledSup = addStyle("sup")`,
+        },
```
kevinb-khan, line 51: suggestion: It might be worth handling things like `addStyle("foo-bar")` to enable handling of custom HTML components.  I know we don't use any right now, but we might someday in the future or other people/orgs might use this eslint rule (it is open source).  Being pro-active is pretty low cost in this situation.

packages/eslint-plugin-khan/test/rules/aphrodite-add-style-variable-name.test.ts
```diff
@@ -0,0 +1,120 @@
...
+});
+
+const ruleName = "aphrodite-add-style-variable-name";
+const rule = rules[ruleName];
+
+ruleTester.run(ruleName, rule, {
+    valid: [
+        {
+            code: `const StyledDiv = addStyle("div")`,
+        },
```
kevinb-khan, line 27: Does this rule ignore things like `const MyCompWithStyle = addStyle(MyComp)`?

kevinb-khan (approved):
LGTM 🎉  Thanks for adding this rule.  I've left a few style suggestions in the comments.

packages/eslint-plugin-khan/demo/.eslintrc.js
```diff
@@ -1,6 +1,7 @@
-/* eslint-disable import/no-commonjs */
 module.exports = {
+    root: true,
```
beaesguerra, line 2: We set this so that the demo eslint config isn't merged with parent eslint config files!

packages/eslint-plugin-khan/demo/src/foo.tsx
beaesguerra: Circling back so I remember what we did:
- We updated eslint versions
- We updated the demo project to use the released `eslint-plugin-khan` package. We'll enable the new addStyle rule in the demo project once these changes are released
- We ignored the demo project from lint checks so that CI can pass (it is expected that the demo project has lint errors so we can test that rules are working as expected!)

packages/eslint-plugin-khan/test/rules/aphrodite-add-style-variable-name.test.ts
beaesguerra: Good question, yes! It will return early if the addStyle argument is not a string:

https://github.com/Khan/wonder-stuff/blob/fc49725b62cbb3fb7cfb179c5eb71e6d7f3fe323/packages/eslint-plugin-khan/src/rules/aphrodite-add-style-variable-name.ts#L79-L85

I updated the tests and docs to explain that the naming convention only applies to string arguments for html tags!

packages/eslint-plugin-khan/test/rules/aphrodite-add-style-variable-name.test.ts
beaesguerra: Good idea! Instead of capitalizing the tag name for the expected variable name, I updated it to turn the string into PascalCase. I added a test for this!

packages/eslint-plugin-khan/src/rules/aphrodite-add-style-variable-name.ts
beaesguerra: Good idea! I moved it to a `toPascalCase` function to handle the custom html element use case you mentioned!

packages/eslint-plugin-khan/src/rules/aphrodite-add-style-variable-name.ts
beaesguerra: Thanks for the feedback Kevin! I refactored the function to favour early returns. It is easier to read now 😄

packages/eslint-plugin-khan/docs/aphrodite-add-style-variable-name.md
```diff
@@ -0,0 +1,35 @@
+# Naming convention for addStyle variable (aphrodite-add-style-variable-name)
```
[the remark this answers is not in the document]
kevinb-khan: Thanks for following the existing patterns.  LGTM :)

packages/eslint-plugin-khan/test/rules/aphrodite-add-style-variable-name.test.ts
kevinb-khan: I don't see any uses of it in our code.  I usually tell people to manually add a `style` prop.  Since that's what all of our code does I guess people just copy/pasting that now which is good.

packages/eslint-plugin-khan/test/rules/aphrodite-add-style-variable-name.test.ts
kevinb-khan: Awesome. Thank you!

kevinb-khan (approved):
Thanks for all of the changes.  This looks great.  🚀

PR #1096, opened 2024-12-13
Update error message for aphrodite-add-style-variable-name rule

beaesguerra:
## Summary:
Updating the error message so it is more useful (noticed this after testing with the demo project)

Previously: `Variable name "div" does not match tag name "div". Variable name should be "StyledDiv"`

Screenshot 2024-12-13 at 2 26 27 PM

New: `Variable name "div" does not match the expected naming convention. Expected: "StyledDiv"` 

Issue: FEI-5952

## Test plan:
- Tests continue to pass
- Will integrate with the demo project once this update is released!

### DIFF
--- a/.changeset/perfect-cycles-fry.md
+++ b/.changeset/perfect-cycles-fry.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/eslint-plugin": patch
+---
+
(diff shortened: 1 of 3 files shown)


kevinb-khan (approved):
Makes sense.  Thanks for fixing that. 🚢

PR #1114, opened 2025-02-03
Create eslint config for a11y

beaesguerra:
## Summary:
This work is part of the implementation of [ADR#781 Enabling more lint rules for accessibility](https://khanacademy.atlassian.net/wiki/x/IoBVyg)

In these changes, a separate eslint config file for accessibility is set up since not all repos will need a11y config (see [config thread](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/3394600994/ADR+781+Enabling+more+lint+rules+for+accessibility?focusedCommentId=3428155481) from the ADR for more details). Following the [eslint docs for Sharing Multiple Configs](https://eslint.org/docs/latest/extend/shareable-configs-deprecated#sharing-multiple-configs), it is consumed in projects by adding the following to the project's eslint config file:
```js
extends: ["@khanacademy/eslint-config/a11y"]
```

The a11y config file currently enables the `aphrodite-add-style-variable-name` rule from `@khanacademy/eslint-plugin`. This is the first step to adding more a11y linting rules so that the naming convention for html elements is more predictable for [custom component mapping](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y#component-mapping) for `eslint-plugin-jsx-a11y` (see [rule docs](https://github.com/Khan/wonder-stuff/blob/main/packages/eslint-plugin-khan/docs/aphrodite-add-style-variable-name.md) for more details).

Next steps:
- Use this new config in WB, perseus, webapp and address lint errors from the aphrodite-add-style-variable-name rule
- Update the a11y.js config with the config based on the accessibility linting rules ADR
- Use the updated config in projects and address existing errors by disabling the rules per line. Teams can address existing errors as they work in the area and new errors can be prevented with these lint rules :) 

Issue: FEI-6050

## Test plan:
Testing extending the new config locally:
- Run `yarn pack` in `packages/eslint-config-khan`. This will create a `.tgz` file in the directory
- In another project (like Wonder Blocks), install the package: ex: `yarn add -D -W /Users/beaesguerra/khan/wonder-stuff/packages/eslint-config-khan/khanacademy-eslint-config-v5.0.1.tgz`
  - Note: you might need to remove the `node_modules` first to make sure it installed the correct package. You can check by checking if the file exists at `node_modules/@khanacademy/eslint-config/a11y.js` in the project
  - Make sure `@khanacademy/eslint-plugin` is at v3.1.1
- Add `"@khanacademy/eslint-config/a11y"` to the `extends` property in the eslint config file
- Run `yarn lint` in the project and restart the ESLint server. It should show `aphrodite-add-style-variable-name` errors
image
image


packages/eslint-config-khan/a11y.js
```diff
@@ -0,0 +1,27 @@
...
+        // TODO(csilvers): once we properly use node.js for node
+        // files, get rid of this next line.
+        node: true,
+        browser: true,
+        es6: true,
+        jest: true,
+    },
+    settings: {
+        react: {
+            version: "detect",
```
beaesguerra, line 17: Updated the settings for the react version following what we have in `webapp`: https://github.com/Khan/webapp/blob/81dea05cb7a6534847a43e6b057cef0431f60b8c/services/static/.eslintrc.js#L928-L930

packages/eslint-config-khan/a11y.js
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable import/no-commonjs */
```
beaesguerra, line 1: This file is mostly copied from the existing config: https://github.com/Khan/wonder-stuff/blob/main/packages/eslint-config-khan/index.js . I'll leave comments in the PR for what's different!

packages/eslint-config-khan/a11y.js
```diff
@@ -0,0 +1,27 @@
...
+        react: {
+            version: "detect",
+        },
+        "import/parsers": {
+            "@typescript-eslint/parser": [".ts", ".tsx"],
+        },
+        "import/extensions": [".js", ".jsx", ".ts", ".tsx"],
+    },
+    rules: {
+        "@khanacademy/aphrodite-add-style-variable-name": ERROR,
```
beaesguerra, line 25: Rules are different from the main eslint config!

packages/eslint-config-khan/a11y.js
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable import/no-commonjs */
+const ERROR = "error";
+
+module.exports = {
+    parser: "@typescript-eslint/parser",
+    plugins: ["@khanacademy"],
+    env: {
+        // TODO(csilvers): once we properly use node.js for node
+        // files, get rid of this next line.
+        node: true,
```
kevinb-khan, line 10: I think we can get rid of this since we won't using React components in node scripts.

packages/eslint-config-khan/a11y.js
```diff
@@ -0,0 +1,27 @@
...
+        browser: true,
+        es6: true,
+        jest: true,
```
kevinb-khan, line 13: I think we can get rid of this as well since we're importing jest "globals" from `@jest/globals` now.

kevinb-khan (approved):
LGTM.  I left a couple of suggestions to tidy up the the config in inline comments.  Thanks for creating this new config.  I'm excited that our eslint config is becoming more modular. 🎉

packages/eslint-config-khan/a11y.js
beaesguerra: Good call, I've removed the unnecessary env settings, thanks for the suggestions!

PR #1118, opened 2025-02-05
Enable rules from `eslint-plugin-jsx-a11y` in `@khanacademy/eslint-config/a11y`

beaesguerra:
## Summary:
This work is part of the implementation of [ADR#781 Enabling more lint rules for accessibility](https://khanacademy.atlassian.net/wiki/x/IoBVyg)

These changes includes:
- Updating the `eslint-plugin-jsx-a11y` dependency
- Configuring rules and settings (including custom component mapping, attribute mapping, polymorphic components) for `eslint-plugin-jsx-a11y` based on the [proposed rules and settings in the ADR](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/3394600994/ADR+781+Enabling+more+lint+rules+for+accessibility#Proposed-customization-and-overrides-for-the-plugin) and the [POC](https://github.com/Khan/webapp/pull/26840/files#diff-20e55c78ab86b0eb87c6f756500b87f451b52fab251d6f327ce572ec88e8ae3b)
- Updating the README to include instructions for using `@khanacademy/eslint-config/a11y`

Issue: FEI-1133

Implementation Plan:
- https://github.com/Khan/wonder-stuff/pull/1114 Set up eslint a11y config and enable aphrodite-add-style-variable-name 
- Use this new config in WB, perseus, webapp and address lint errors from the aphrodite-add-style-variable-name rule
  - https://github.com/Khan/wonder-blocks/pull/2459
  - https://github.com/Khan/perseus/pull/2193
  - https://github.com/Khan/webapp/pull/29212
- (this PR) Update the a11y.js config with the config based on the accessibility linting rules ADR
- Use the updated config in projects and address existing errors by disabling the rules per line. Teams can address existing errors as they work in the area and new errors can be prevented with these lint rules!

## Test plan:
Testing the new config locally:
- Run yarn pack in `packages/eslint-config-khan`. This will create a `.tgz` file in the directory
- In another project, install the package: ex: `yarn add -D -W /Users/beaesguerra/khan/wonder-stuff/packages/eslint-config-khan/khanacademy-eslint-config-v5.1.0.tgz`
  - Note: you might need to remove the node_modules first to make sure it installed the correct package. You can check by checking `node_modules/@khanacademy/eslint-config/a11y.js` in the project to see if the new config is there
- Make sure the project already extends `@khanacademy/eslint-config/a11y` in the project's eslint config file
- Run `yarn lint`/`pnpm lint` in the project and restart the ESLint server. It should show errors from the `jsx-a11y` plugin

Note: Before merging and releasing these changes, I'll test these changes locally with the different projects and evaluate the errors to make sure the config changes are still relevant and helpful!


packages/eslint-config-khan/README.md
```diff
@@ -30,10 +31,7 @@ For monorepos the `"import/resolver"` settings will look like this:
     },
 ```
 
-For regulard repos, the settings will look like this:
-
-```
-For monorepos the `"import/resolver"` settings will look like this:
```
beaesguerra, line 36: This was duplicated from line 12 and not needed (I think!)

marcysutton (approved):
This looks great to me, thanks for all the thorough comments and rule settings! It will be good to hear @kevinb-khan's take on it too.

packages/eslint-config-khan/a11y.js
```diff
@@ -16,8 +26,100 @@ module.exports = {
...
+                StyledPre: "pre",
+                StyledSup: "sup",
+                StyledMark: "mark",
+                StyledTable: "table",
+                StyledTr: "tr",
+                StyledTd: "td",
+                StyledTh: "th",
+                StyledDl: "dl",
+                StyledDt: "dt",
+                StyledDd: "dd",
```
kevinb-khan, line 70: So many.  It would be nice to have a wonder-blocks package that exports this.  Is that something that's on the roadmap?

packages/eslint-config-khan/a11y.js
```diff
@@ -1,9 +1,19 @@
 /* eslint-disable import/no-commonjs */
 const ERROR = "error";
+const OFF = "off";
```
kevinb-khan, line 3: I wondering if switching to `tseslint.config` means we'll get property type checking for our configs.  No action req'd, just something we may want to think about for the future.  See https://typescript-eslint.io/packages/typescript-eslint#config for details.

kevinb-khan (approved):
LGTM.  I didn't look at the specific rules though.  🎉

packages/eslint-config-khan/a11y.js
beaesguerra: Thanks for sharing, that would be helpful down the line, yes! I was following the format from our default config file 😅

packages/eslint-config-khan/a11y.js
beaesguerra: Do you mean a WB package that exports this StyledX component mapping? Would it be useful for other cases? If not, another option would be to move this mapping to another file in this package and import it from there so the config file isn't as long! What do you think?

Or did you mean it'd be nice to have a WB package that exports the StyledX components already (as we briefly [chatted](https://khanacademy.slack.com/archives/C8Z9DSKC7/p1737662594706209) about 😄)?

beaesguerra:
I tested the config changes locally in `webapp`, `wonder-blocks` and `perseus` and evaluated the errors! I found a few more ways to configure the settings so that the lint rules are more effective, the config is updated now (see last 2 commits) 😄

packages/eslint-config-khan/a11y.js
```diff
@@ -58,6 +58,12 @@ module.exports = {
                 IconButton: "button",
                 TextArea: "textarea",
                 TextField: "input",
+                // SingleSelect/MultiSelect components don't have an equivalent
+                // html element, the closest is a `select` element. This allows
+                // for labels to be assosciated with the control for the
+                // jsx-a11y/label-has-associated-control rule.
+                SingleSelect: "select",
+                MultiSelect: "select",
```
marcysutton, line 66: Nice. Good find!!

marcysutton (approved):
This is awesome!! I tested it out locally in Wonder Blocks and it worked great. One minor change in the repro steps was that I had to use lowercase `-w` with pnpm. 

Once I got that and the correct tarball path, it installed properly and caught non-semantic elements with `addStyle`. Great work, @beaesguerra!

packages/eslint-config-khan/a11y.js
kevinb-khan: I was thinking of a WB package that does a bunch of exports like:
```ts
export const StyleP = addStyle("p");
export const styledTable = addStyle("table");
...
```
This suggestion is only tangentially related to this PR.  It's so that people don't need to remember to use `addStyle` to get these variants.  Eventually we could even add lint rules to disallow non-styled variants.

beaesguerra: Thanks for clarifying! I created https://khanacademy.atlassian.net/browse/WB-1879 for this 😄

PR #1122, opened 2025-02-28
Update wonder-stuff-sentry to Sentry v9

jeremywiebe:
## Summary:

This PR updates the `@khanacademy/wonder-stuff-sentry` package to be compatible with `@sentry/browser` v9. The major difference is that in v8 and later, integrations are no longer classes, simply objects. The pattern I saw in other samples is to write a function that initializes the Sentry integration with whatever options provided. 

The other big difference is that we no longer need to use `addGlobalEventProcessor`. In fact, that function no longer exists. In v8 though we can simplify the integration considerably by implementing the `processEvent` hook on the `Integration`. This accomplishes pretty much exactly what the `setupOnce()` logic was doing previously (see https://github.com/getsentry/sentry-javascript/blob/5060542296ead088486e7cc2472508a595a7190b/packages/core/src/integration.ts#L121-L126). 

Issue: FEI-6242

## Test plan:

I plan to integrate this into a consuming app and test it there to make sure the integration works. Also, all existing tests passed after updating them to the new integration (or deleting tests which no longer applied - ones relating to integration setup).


packages/wonder-stuff-sentry/package.json
```diff
@@ -27,12 +27,18 @@
...
     },
     "devDependencies": {
-        "@khanacademy/ws-dev-build-settings": "^2.0.1"
+        "@khanacademy/ws-dev-build-settings": "^2.0.1",
+        "@sentry/browser": "8",
+        "@sentry/core": "8"
+    },
+    "peerDependencies": {
+        "@sentry/browser": "8",
+        "@sentry/core": "8"
```
somewhatabstract, line 36: **question:** I'm surprised these weren't already peer deps. My bad. Also, usually if something is in `peer` it should automatically installed for dev too. But it can't hurt to be explicit.

However, why not just jump straight to 9?

packages/wonder-stuff-sentry/src/kind-error-data.ts
somewhatabstract, line 1: **suggestion[change-required]:** Please rename the file to match the export.

packages/wonder-stuff-sentry/src/__tests__/kind-error-data.test.ts
somewhatabstract, line 1: **suggestion[change-required]:** Please rename the file to match the file under test.

packages/wonder-stuff-sentry/package.json
somewhatabstract: **suggestion:** Consider using the `@sentry/types` dependency to provide types. It's more comprehensive than just the types from the regular dependencies.

somewhatabstract (changes requested):
Thanks for taking this on. See the epic I raised (FEI-6239) for this stuff. I think we should go straight to v9 and skip v8.

Some changes requested inline but just housekeeping really.

somewhatabstract (approved):
Thanks for taking this on. See the epic I raised (FEI-6239) for this stuff. I think we should go straight to v9 and skip v8.

Some changes requested inline but just housekeeping really.

(Sorry for the double review post; finger slipped and hit request changes and I hit the other button before it finished...seems to have done both :D )

packages/wonder-stuff-sentry/package.json
jeremywiebe: I figured it'd be more principled to go through 8 first. But if you are good with going straight to 9, that's what I'll do in this PR. :)

jeremywiebe: In v8 and newer, the types needed have been moved to `@sentry/core` and deprecated from `@sentry/types`. 

image

somewhatabstract: Great! I think they made that transition during v8.

packages/wonder-stuff-sentry/src/__tests__/__snapshots__/kind-error-data-integration.test.ts.snap
somewhatabstract, line 1: **question:** I would expect this to be a moved file also? Is the other snapshot not deleted?

jeremywiebe: Good catch! Fixed that.

PR #1123, opened 2025-02-28
Migrate to pnpm

jeremywiebe:
## Summary:

This PR migrates this repo to [pnpm](pnpm.io). After the initial import, there were several changes I had to make to get all the `scripts` working. Many of these were the result of `yarn` resolving modules through the workspace root's `node_modules` whereas `pnpm` puts a `node_modules` dir in each package's dir.

_**Note**: the cache priming action is failing because it uses the action definition from the base branch (ie. `main`) which still references `yarn`, while the branch it's trying to merge has a `package.json` that has a `packageManager` entry that specifies `pnpm`. This should be resolved once we land this branch._

Changes include: 
  * Use workspace:* version for all in-repo dependencies
  * Prevent usage of package managers other than pnpm and pin pnpm to v10.0.0
  * Approve packages that have post-install scripts (pnpm doesn't run a script's `post-install` script unless it's approved) #security
  * Move all scripts to use pnpm instead of yarn
  * Fix globbing in `check-type-definitions.js` so that `pnpm build:types` works
  * Fix globbing in `pre-publish-check-ci.js`
  * ESLint: ignore generated `docs/` directory
  * Add dependencies to `eslint-plugin-khan` that used to be available implicitly by yarn
  * ESLint: allow `typedoc.config.js` to be CJS
  * Migrate typedoc config to be a js file so we can just glob the packages to build docs for

Issue: FEI-6300

## Test plan:

I ran each of the scripts in `package.json` including the `publish:ci` (which I temporarily modified to not do the last step which acctually publishes).


package.json
```diff
@@ -74,28 +72,37 @@
     "typescript": "5.7.3",
     "winston": "^3.10.0"
   },
-  "resolutions": {
-    "//": "we need to pin wide-align/string-width because v5 & up are ESM only",
-    "wide-align/string-width": "^4.2.0",
-    "node-gyp": "^10.0.0"
+  "pnpm": {
+    "overrides": {
+      "a": "we need to pin wide-align/string-width because v5 & up are ESM only",
```
jeremywiebe, line 77: `pnpm` was complaining about the `//` resolutions when I migrated to `overrides`. Using an alphanumeric package name for the comment appears to work but feels pretty gross. 

pnpm does support a package.json that includes comments (`package.json5`) but from what I've read not all other JS tooling does so that feels like a deal-breaker. 

Do we still need these overrides at all?

packages/eslint-plugin-khan/package.json
```diff
@@ -14,10 +14,17 @@
   "module": "dist/es/index.js",
   "author": "Kevin Barabash <kevinb@khanacademy.org>",
   "license": "MIT",
+  "dependencies": {
+    "ancesdir": "^5.0.1",
+    "@babel/types": "^7.23.0",
+    "@typescript-eslint/utils": "8.17.0",
+    "checksync": "^5.0.5"
+  },
```
jeremywiebe, line 22: These were moved up from below `devDependencies`, but also added the `ancesdir` dep. It used to work with `yarn` but `pnpm` lays out the `node_modules` dir differently and so we no longer see another package's (or root's) dependencies and we need to be explicit.

jeremywiebe:
The Lint workflow is failing because the compressed-size-action is broken when the current and base branch use different package managers. This is a temporary issue and I don't know that I want to work through upstreaming a fix for this. 

Unless folks disapprove, I plan to land this PR without addressing that issue. Future PRs won't have to worry about that because they won't be switching between package managers.

build-settings/check-type-definitions.ts
```diff
@@ -24,6 +24,8 @@ const packagesDir = path.join(rootDir, "packages");
 const typeDefinitionFiles = fglob.sync("**/*.d.ts", {
     cwd: packagesDir,
     onlyFiles: true,
+    globstar: true,
+    ignore: ["**/node_modules/**"],
```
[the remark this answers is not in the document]
somewhatabstract: Should this be `**/packages/**/node_modules/**`?

Are the types being generated correct?

packages/wonder-stuff-render-environment-jsdom/examples/README.md
somewhatabstract, line 1: **suggestion:** You can delete this whole package. See ADR-825 and FEI-6388. I was trying to get it done before you had to migrate.

packages/wonder-stuff-testing/package.json
```diff
@@ -17,21 +17,21 @@
     "main": "dist/index.js",
     "types": "dist/index.d.ts",
     "scripts": {
-        "test": "bash -c 'yarn --silent --cwd \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
+        "test": "bash -c 'pnpm --silent --dir \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
```
somewhatabstract, line 20: **suggestion:**
```suggestion
        "test": "bash -c 'pnpm run --silent --dir \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
```

packages/wonder-stuff-render-server/package.json
```diff
@@ -18,14 +18,14 @@
     "types": "dist/index.d.ts",
     "scripts": {
         "example": "bash -c 'RUNNER=\"./examples/$0/run.ts\"; KA_LOG_LEVEL=silly NODE_ENV=development babel-watch --config-file ../../babel.config.js \"$RUNNER\" --extensions .ts'",
-        "test": "bash -c 'yarn --silent --cwd \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
+        "test": "bash -c 'pnpm --silent --dir \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
```
somewhatabstract, line 21: **suggestion:**
```suggestion
        "test": "bash -c 'pnpm run --silent --dir \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
```

packages/wonder-stuff-sentry/package.json
```diff
@@ -21,13 +21,13 @@
         "CHANGELOG.md"
     ],
     "scripts": {
-        "test": "bash -c 'yarn --silent --cwd \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
+        "test": "bash -c 'pnpm --silent --dir \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
```
somewhatabstract, line 24: **suggestion:**
```suggestion
        "test": "bash -c 'pnpm run --silent --dir \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
```

packages/wonder-stuff-server/package.json
```diff
@@ -17,13 +17,13 @@
     "main": "dist/index.js",
     "types": "dist/index.d.ts",
     "scripts": {
-        "test": "bash -c 'yarn --silent --cwd \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
+        "test": "bash -c 'pnpm --silent --dir \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
```
somewhatabstract, line 20: **suggestion:**
```suggestion
        "test": "bash -c 'pnpm run --silent --dir \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
```

utils/pre-publish-check-ci.js
```diff
@@ -14,7 +14,7 @@ const {
 } = require("./pre-publish-utils");
 
 // eslint-disable-next-line promise/catch-or-return
-fg(path.join(__dirname, "..", "packages", "*", "package.json")).then(
+fg("packages/*/package.json", {cwd: path.resolve(__dirname, "..")}).then(
```
somewhatabstract, line 17: **suggestion:** I feel like `ancesdir` would simplify this a bit.
```suggestion
fg("packages/*/package.json", {cwd: ancesdir()}).then(
```

Then also if the file moves, it doesn't need editing to get the path right.

packages/wonder-stuff-testing/package.json
somewhatabstract, line 1: **note:** This package needs deleting too but it's gnarly because of the wonder-stuff-ci dependency. So don't bother trying. I'll tackle it.

packages/wonder-stuff-render-server/examples/README.md
somewhatabstract, line 1: **suggestion:** You can delete this whole package. See ADR-825 and FEI-6388. I was trying to get it done before you had to migrate.

somewhatabstract (approved):
Thanks for tackling this. I think there are some edits needed (see inline comments), but I'll leave that to your discretion.

package.json
```diff
@@ -74,28 +72,37 @@
...
+    "publish:ci": "pnpm run lint:pkg-json && node utils/pre-publish-check-ci.js && git diff --stat --exit-code HEAD && pnpm run build && pnpm run build:types && changeset publish",
+    "test": "pnpm jest",
     "typecheck": "tsc",
-    "nochangeset": "yarn changeset add --empty",
-    "add:devdepbysha": "bash -c 'yarn add -W --dev \"git+https://git@github.com/Khan/$0.git#$1\"'"
-  }
+    "nochangeset": "pnpm changeset add --empty",
+    "add:devdepbysha": "bash -c 'pnpm add --workspace-root --save-dev \"git+https://git@github.com/Khan/$0.git#$1\"'"
+  },
+  "packageManager": "pnpm@10.0.0+sha512.b8fef5494bd3fe4cbd4edabd0745df2ee5be3e4b0b8b08fa643aa3e4c6702ccc0f00d68fa8a8c9858a735a0032485a44990ed2810526c875e416f001b17df12b"
```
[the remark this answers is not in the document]
jeremywiebe: This is the exact value that we pin `pnpm` to in other projects. I think if we want to move to a newer one that'd be a good thing to do in a separate PR (and perhaps just churn through all repos we've migrated to `pnpm` in one shot).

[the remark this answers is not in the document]
somewhatabstract: Yeah, I know, but just copying it seems...I don't know. What's the reason for pinning it for everything. The tooling (corepack, etc.) (at least for me with my different repos) handles the version changes seamlessly.

I don't think propagating a pinned version is necessary.

build-settings/check-type-definitions.ts
```diff
@@ -24,6 +24,8 @@ const packagesDir = path.join(rootDir, "packages");
 const typeDefinitionFiles = fglob.sync("**/*.d.ts", {
     cwd: packagesDir,
     onlyFiles: true,
+    globstar: true,
```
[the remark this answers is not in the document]
jeremywiebe: It's not. Thanks for catching this. I set it explicitly while trying to figure out why it was breaking, but the next line (`ignore`) is what fixed it.

package.json
jeremywiebe: I'm leaning that way too! Git should have details as to why the line was added even if it takes a bit of digging.

somewhatabstract: Could have a separate section for commenting that is custom of `packageName -> comment` and then have a test that verifies there's an override per comment and vice versa? Seems like a lot of work for now...so still...just delete them?

somewhatabstract: > I'm leaning that way too! Git should have details as to why the line was added even if it takes a bit of digging.

This is harder than you might think and generally, doesn't happen unless someone absolutely needs to know. We have had this issue in the past quite a bit (why was this resolution there? Let me just get rid of it. Oh I can fix this error...) wasting lots of time when a comment right there would've said "we tried all that, just do this...see this ticket for info". Finding the right commit in Git can be a pain sometimes as code changes and moves around. Having the comments colocated was super useful and time-saving.

package.json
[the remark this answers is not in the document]
jeremywiebe: One reason is that pinning to a version with a hash like this gives us some protection against a supply chain attack. I would agree that there might not be much use just for that. 

Happy to move this to be the most recent version. I do think it's valuable to do it the same way across our Khan Academy repos though.

packages/wonder-stuff-ci/package.json
```diff
@@ -17,16 +17,16 @@
     "main": "dist/index.js",
     "types": "dist/index.d.ts",
     "scripts": {
-        "test": "bash -c 'yarn --silent --cwd \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
+        "test": "bash -c 'pnpm --silent --dir \"../..\" test ${@:0} $($([[ ${@: -1} = -* ]] || [[ ${@: -1} = bash ]]) && echo $PWD)'"
```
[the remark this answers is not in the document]
jeremywiebe: I've added it. Still doesn't run just the tests in the current package.  I'm going to leave these here for now.

packages/wonder-stuff-ci/package.json
```diff
@@ -17,16 +17,16 @@
...
     },
     "devDependencies": {
         "@types/node": "^20.17.23",
-        "@khanacademy/ws-dev-build-settings": "^2.0.1"
+        "@khanacademy/ws-dev-build-settings": "workspace:"
```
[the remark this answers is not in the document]
jeremywiebe: Yes! Thanks for catching that.

package.json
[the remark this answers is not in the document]
somewhatabstract: Yes, pinning for sure. I think that it's demanded (all the repos I updated rely on that), I'm questioning pinning to the same version as everything else.

package.json
jeremywiebe: Ok, I brought it back with a pattern I think works. Install works which I think is the only place that'd parse that section of `package.json`. I do wonder if our current deps need that override, but I'm paranoid to remove it and how to fully verify.

packages/wonder-stuff-render-environment-jsdom/examples/README.md
jeremywiebe: I would prefer to avoid doing that as part of this PR especially given this is focused on pnpm. :)

jeremywiebe:
I can't find the comment anymore, but in reply to the question about whether types build properly, they do. I build them on `main` and then on this branch and compared the sets of `dist` folders. 

They were identical except for extremely small differences in the sourcemaps `mappings` field.

package.json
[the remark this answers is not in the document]
jeremywiebe: Ok, had a chat with John and I've changed my mind. I'm going to just change this to be `pnpm@10.6.5` (latest). We can change that in a future PR if we want.

PR #1137, opened 2025-03-20
[🔥AUDIT🔥] Update 'cleanup' script to remove _all_ node_modules folders

jeremywiebe:
## Summary:

I am blocked from landing Khan/wonder-stuff#1123 because the `clean` step does not remove all `node_modules` directories. We use the `preactjs/compressed-size-action` to check build size changes in each PR. This action switches to the upstream branch after building on the PR branch. It then issues a 'clean' command (which we defined in our `package.json`). Finally, it installs packages and builds on the upstream branch.

When a PR (#1123) switches package managers and leaves some `node_modules` folders in place after cleaning we have issues. In this case`yarn` ttempts to install over a `./node_modules` that was set up by `pnpm`, and [fails](https://github.com/Khan/wonder-stuff/actions/runs/13980489827/job/39144501346?pr=1123) (see screenshot below):

image

Issue: --none--

## Test plan:

I ran the following steps which simulate what's happening in the Check Builds step of PR #1123:

  1. `git co jer/pnpm`
  2. `find . -name node_modules -prune -exec rm -rf {} \;` # make sure _all_ node_modules are deleted
  3. `pnpm install`
  4. `git co jer/fix-cleanup`
  5. `pnpm clean` # yes, using `pnpm` as that's how the `preactjs/compressed-size-action` does it
  6. `yarn install` # this is the step that fails in the GitHub Action 

Success.


package.json
```diff
@@ -86,7 +86,7 @@
     "build:types": "yarn tsc --build --verbose tsconfig-build.json && ./build-settings/check-type-definitions.ts",
     "build:docs": "typedoc",
     "watch": "yarn rollup --watch",
-    "clean": "rm -rf packages/wonder-stuff-*/dist && rm -rf packages/wonder-stuff-*/node_modules && rm -f packages/*/*.tsbuildinfo",
+    "clean": "rm -rf packages/wonder-stuff-*/dist && rm -rf node_modules && rm -rf packages/*/node_modules && rm -f packages/*/*.tsbuildinfo",
```
somewhatabstract, line 89: **suggestion:** Hmm, I don't think this is what we want. Folks use this locally to clean too and it would be a pain if it obliterated the working state every time.

Can you add a new `clean:all` script or something and have the appropriate workflow call that instead?

somewhatabstract: NVM, digging deeper, I think just keep what you have. We can revisit if we need to.

somewhatabstract (approved):
LGTM

somewhatabstract (approved):
LGTM

PR #1139, opened 2025-03-21
Mark demo project as non-publishable

jeremywiebe:
## Summary:

The most recent release failed near the end because it tried to publish the `demo` project. This PR changes the demo project to be explicitly private. This will help changesets know not to try to publish it. 

This didn't happen in the demo project in earlier releases because it wasn't in the Yarn workspace. With pnpm, we can add it to the workspace, but that caused changesets to try to publish it.  By marking it as `private: true`, we'll block that in the future. 

image

Issue: FEI-6300

## Test plan:

We'll need to land this and then release the Version Packages PR that's created by it.

### DIFF
--- a/.changeset/happy-dots-bake.md
+++ b/.changeset/happy-dots-bake.md
@@ -1,0 +1,5 @@
+---
+"demo": patch
+---
+
+Mark demo project as private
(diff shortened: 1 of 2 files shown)


somewhatabstract (approved):
Darn, sorry I missed this in the earlier reviews. Thanks for making sure this is all good.

PR #1238, opened 2026-04-02
Update workflows to use versions of actions that use Node v24

kevinb-khan:
## Summary:
Node v20 is exiting its LTS window at the end of April and GitHub is requiring all workflows/actions to use Node v24.  Because of this, we need to make sure all of the actions we're using are on v24.

Issue: FEI-7601

## Test plan:
- let CI run and ensure that all checks pass without any warnings about Node v20

Screenshot 2026-04-02 at 5 01 06 PM

I've copy/past'd the one remaining warning below.  I think it's safe to ignore this warning.
> Starting April 2026, the CodeQL Action will skip computing file coverage information on pull requests to improve analysis performance. File coverage information will still be computed on non-PR analyses.
> 
> To opt out of this change, set the `CODEQL_ACTION_FILE_COVERAGE_ON_PRS` environment variable to `true`. Alternatively, create a custom repository property with the name `github-codeql-file-coverage-on-prs` and the type "True/false", then set this property to `true` in the repository's settings.


jeresig (approved):

somewhatabstract (approved):
Great. Thanks for doing this.

Aside: I think we should probably get things working with SHAs instead of tags in Wonder Stuff at some point.

PR #1241, opened 2026-04-28
Update ts/js settings for new version of VSCode extension

jeremywiebe:
## Summary:

I noticed that a bunch of settings in `.vscode/settings.json` were underlined with a yellow squiggly line (see screenshot). In recent releases of the TypeScript/JavaScript extension for VSCode, they've unified settings from being prefixed with separate values (`javascript.*` and `typescript.*`) to a single unified namespace (namely `js/ts.*`).

This PR updates our settings to use this new, unified namespace.

| Before | After |
| --- | --- |
|  |  |

_**Note**: The screenshots above are from Perseus so the settings may be _slightly_ different_

Issue: LEMS-4064

## Test plan:

Open up the project in VSCode and note that the same set of inlay hints are still enabled.


jeresig (approved):

packages/wonder-stuff-logging/src/types.ts
```diff
@@ -1,14 +1,14 @@
...
+    Logform,
     transport as Transport,
 } from "winston";
 import {Runtime} from "@khanacademy/wonder-stuff-core";
 
 /**
  * Describes logging metdata.
  */
-export type Info = LogEntry;
+export type Info = Logform.TransformableInfo;
```
jeremywiebe, line 11: These were broken by a Dependabot bump in https://github.com/Khan/wonder-stuff/pull/1204

PR #1244, opened 2026-07-20
Upgrade repo to Node v24.18.0

jeremywiebe:
## Summary:

There are two high-severity CVE's for Node v24.16.0. We're skipping v24.17.0 because we tried it previously in the Perseus repo and it broke one of our Github Actions.

This PR also applies updated `agent-settings` (given that we bumped the Node version in some of the hooks in that dir).

Issue: INFRA-11108

## Test plan:

Run all Github Actions and if they all pass, we should be good


kevinb-khan (approved):
LGTM, but you should probably wait on landing this until the khan-dotfiles change lands and you've waited a couple of days for OLC's auto-updated to do its thing.

kevinb-khan:
Thanks for updating node in this repo. 🎉

.claude/hooks/echo-banner-filter.mjs
jeremywiebe, line 1: ALL of the changes to the AI agent directories are the result of running: `pnpm install-settings --env all wonder-stuff` in the Khan/agent-settings directory after [bumping](https://github.com/Khan/agent-settings/pull/36) it to Node 24.18.

jeremywiebe:
@kevinb-khan Sorry, requesting one more pass on this PR. The changes since you reviewed are just applying the updated agent-settings files after landing https://github.com/Khan/agent-settings/pull/36

somewhatabstract (approved):
@jeresig @kevinb-khan @jeremywiebe what do you think about using a .gitattributes file to mark some of the AI files as generated so we don't get them expanded in these PRs that update them?

jeresig:
@somewhatabstract sure! that sounds like a good idea - maybe I can update the agent-settings logic to set that up automatically for us.

.gitattributes
```diff
@@ -1 +1,33 @@
 pnpm-lock.yaml linguist-generated
+
+# BEGIN: agent-settings managed files
```
jeremywiebe, line 3: I ran `pnpm install-settings --env all wonder-stuff` in the agent-settings dir and captured all of the files it wrote. I then transformed that list into this list.  

Whenever we automate `agent-settings` writing this list, we at least have comments surrounding this list to key off of so its easy to remove these hand-written entries.

